PR #3, opened 2022-10-20
init structure

WillDudley:


pseudo-rnd-thoughts (commented):
Why do we have code examples from PZ or script files that won't work for kabuki

WillDudley:
Ah, need to delete those scripts and files - thought I did. Thanks!

PR #6, opened 2022-10-25
Wd/mdp dataset

WillDudley:
Adds datastructure for offline datasets

Before merge:
- reinstate precommits
- remove this branch from github workflows


WillDudley:
This PR focusses on the client/server stuff and less so the dataclass - see https://github.com/Farama-Foundation/Kabuki/pull/7 for that

(MDPDataset was used as an example)

PR #31, opened 2023-02-15
Update core API, removing all the C dependencies in favour of python versions

rodrigodelazcano:
# Description
This PR introduces core changes to the Minari API by removing all of the C dependencies and improving the way that users will collect and store new datasets.

Some of the D4RL datasets have been recreated following this PR in this repository:
https://github.com/rodrigodelazcano/d4rl-minari-dataset-generation
These datasets are currently hosted in the Farama servers, check their name with:
```python
# shows the dataset_name of all the Minari datasets available in the Farama server
minari.list_remote_datasets()
```
To download any of the datasets locally to `.minari/datasets/` use:
```python
minari.download_dataset('dataset_name')
```
## HDF5 file structure
The Minari datasets are stored in HDF5 file format :page_facing_up: to leverage the hierarchical organization and divide the data by episode and type. Inside the HDF5 file (named `main_data.hdf5` for every Minari dataset) each episode is represented as a group :file_folder: containing 5 datasets :floppy_disk: present in all Minari datasets (`observations`, `actions`, `terminations`, `truncations`, and `rewards`). Adding an `infos` subgroup is optional, this can contain other datasets or nested subgroups depending on the type of `info` returned in each step.

Furthermore, other subgroups with extra data independent of the Gymnasium step API returns can be added to the episode group. This will be explained in the next section with the usage of the `DataCollectorV0` wrapper and the `StepDataCallback`.

The intrinsic HDF5 file tree structure has the following template:

:page_facing_up: main_data.hdf5
├ :file_folder: episode_id
│  &nbsp;├ :floppy_disk: observations
│  &nbsp;├ :floppy_disk: actions
│  &nbsp;├ :floppy_disk: terminations
│  &nbsp;├ :floppy_disk: truncations
│  &nbsp;├ :floppy_disk: rewards
│  &nbsp;├ :file_folder: infos
│  &nbsp;│  &nbsp;├ :floppy_disk: info datasets
│  &nbsp;│  &nbsp;└ :file_folder: info subgroup
│  &nbsp;│  &ensp; &emsp; └ :floppy_disk: info subgroup dataset
│  &nbsp;└ :file_folder: extra dataset group
│   &ensp; &emsp; └ :floppy_disk: extra datasets
└ :file_folder: next_episode_id

`h5py` is used to read/write the HDF5 files. More on `h5py`: https://docs.h5py.org/en/stable/quick.html

## Data collection
This PR introduces the `DataCollectorV0` wrapper. This is a wrapper over a Gymnasium environment to make the data collection of each environment step agnostic to the user. The initial idea comes from https://github.com/Farama-Foundation/Gymnasium/pull/203 .

The wrapper can be initialized as follows:

```python
import minari
import gymnasium as gym

env = gym.make('EnvID')
env = minari.DataCollectorV0(env, step_data_callback=StepDataCallback, episode_metadata_callback=EpisodeMetadataCallback, record_infos=True, max_buffer_steps=100000, max_buffer_episodes=None)
```
What each argument is for:

- `step_data_callback`: pass an instance of the `StepDataCallback` class, also implemented in this PR. This class is called after each environment step and adapts the step returns to be stored in the internal buffer. `StepDataCallback` returns a dictionary with the data to be added to each dataset in the episode group template of the HDF5 file. Five keys are mandatory : `observations`, `actions`, `terminations`, `truncations`, and `rewards`, and the values must be list or arrays. Nested dictionaries can also be added and this will create an additional subgroup inside the episode group (`{'dataset_name': array/list, 'subgroup_name':{'dataset_name': array/list}}`). 

    By default `StepDataCallback` flattens `Dict` and `Tuple` observation/action spaces, but extra filtering or additional data items can be given to the step data dictionary by overriding the `__call__` method in a child class.
- `episode_metadata_callback`: pass an instance of the `EpisodeMetadataCallback` class, also implemented in this PR. Every time before an episode is saved to an HDF5 file this class is called to add metadata to the episode group and datasets. Generally used to add statistic metrics such as `total_steps`, or `sum`, `mean`, `std`, `min`, `max` to the reward dataset of the episode. Override this class to add other custom metadata metrics to each episode.
- `record_infos`: `True` if the step info value is stored in the buffer
- `max_buffer_steps` and `max_buffer_episodes`: one of both arguments can have an `int` value to specify the number of steps or episodes to be stored in the internal buffer list in-memory of the `DataCollectorV0` environment before dumping and saving it into a temporary HDF5 file. This feature allows for caching when dealing with large datasets. If both arguments are default to `None` the data won't be save to disk until the Minari dataset is created.

## Create Minari dataset
The most straightforward way to create a new Minari dataste in this PR is by using the `create_dataset_from_collector_env()` function as follows:
```python
import minari
import gymnasium as gym

# 1.  Wrap the environment with DataCollectorV0 
env = minari.DataCollectorv0(gym.make('EnvID')

env.reset(123)
#2. Step environment to collect data
for _ in range(100):
    # random policy
    action = env.action_space.sample()
    obs, rew, terminated, truncated, info = env.step(action)
    if terminated or truncated:
        env.reset()

# 3. Create dataset, saved locally in Minari database directory
dataset = minari.create_dataset_from_collector_env(dataset_name='minari-dataset-EnvID', collector_env=env)
```
Every time that the `create_dataset_from_collector_env()` function is called the internal memory of the `collector_env` is cleared and will only be accessible through the Minari dataset.

Another option is to create the data buffer without using `DataCollectorV0`. 
The buffer structure must be a list of dictionary episode buffers. Each episode buffer must contain `observations`, `actions`, `terminations`, `truncations`, and `rewards` with numpy array keys with the number of steps as the size of the first dimension (except the `observations` key which should contain the initial observation as well. An example with zero valued buffers:
 
```python
num_steps = 10
num_episodes = 5

array_shape = (num_steps, 1)

episode = {'observations': np.zeros((num_steps +1, 1)), 'actions': np.zeros(array_shape), 'rewards': np.zeros(array_shape), 'terminations': np.zeros(array_shape), 'truncations': np.zeros(array_shape)}
buffer = [episode for _ in range(num_episodes)]
dataset = minari.create_dataset_from_buffer(buffer)
```
### Update datasets with more data
The Minari datasets can be updated by appending more data to them with the following function:
```python
dataset.update_dataset_from_collector_env(collector_env)
```
This can serve as a checkpoint when collecting and creating a new dataset. An example is given here: https://github.com/rodrigodelazcano/d4rl-minari-dataset-generation/blob/58ad476b5865bf1cbef1afe953e789ca677f7cee/scripts/pointmaze/create_pointmaze_dataset.py#L79

This feature can also be performed directly from list buffers
```python
dataset.update_dataset_from_buffer(buffer)
```

### Combine multiple datasets
Multiple already existent Minari datasets can be combined under a different name as follows:
```python
dataset_v1 = minari.load_dataset('dataset-v1')
dataset_v2 = minari.load_dataset('dataset-v2')

dataset_v3 = minari.combine_datasets([dataset_v1, dataset_v2], 'dataset-v3')
```

## MinariDataset
The `MinariDataset` class makes the datasets metadata accessible and in future contributions the plan is to add sampling and filtering of the episodes.
Another feature in `MinariDataset` is the possibility of recovering the Gymnasium environment the dataset was created with. This is possible thanks to this PR https://github.com/Farama-Foundation/Gymnasium/pull/292 and it can be done as follows:
```python
dataset = minari.load_dataset('dataset_v1')

env = dataset.recover_environment()
```
## 
This PR also modifies how the dataset files are stored. Instead of making a single HDF5 file with the name of the dataset i.e. `dataset_v1.hdf5`, a new directory is created with the name of the dataset and inside it another directory is created named `data` which contains the dataset file `main_data.hdf5`. Other dataset file names can be found in the `data` directory by the name of `additional_data_id.hdf5`. This will be the case if the dataset is updated with `dataset.update_from_collector_env()` (all data files in the `data` directory are linked and when opening `main_data.hdf5` the episode data of every file will be read as together as if it were a single file).

The reasoning behind this directory tree structure is to add a `policies` directory and README's to each dataset folder.

# Next steps
More or less in priority order:
- [ ] Add sampling methods to `MinariDataset` to use the datasets for learning
- [ ] Fix CI
- [ ] Documentation/Update tutorials
- [ ] Evaluate datasets with learning libraries: CORL or d3rlpy
- [ ] New feature to save expert policies (architecture/weights)
- [ ] Generate more datasets. The rest of the D4RL environments that currently have Gymnasium support: MuJoCo, KitchenFranka, and Minigrid
- [ ] Allow for HDF5 compression
- [ ] Import Crowdplay datasets
- [ ] Update `DataCollectorV0` to support vectorized environments
- [ ] Add README to each dataset for description and additional information


pyproject.toml
```diff
@@ -0,0 +1,100 @@
...
+# Some modules are missing type stubs, which is an issue when running pyright locally
+reportMissingTypeStubs = false
+# For warning and error, will raise an error when
+reportInvalidTypeVarUse = "none"
+
+# reportUnknownMemberType = "warning"  # -> raises 6035 warnings
+# reportUnknownParameterType = "warning"  # -> raises 1327 warnings
+# reportUnknownVariableType = "warning"  # -> raises 2585 warnings
+# reportUnknownArgumentType = "warning"  # -> raises 2104 warnings
+reportGeneralTypeIssues = "none"  # -> commented out raises 489 errors
```
pseudo-rnd-thoughts, line 92: We don't need to change in this PR but this should not be "None", should be "Error"

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+from gymnasium import spaces
+
+
+STEP_DATA_KEYS = {
+    "actions",
+    "observations",
+    "rewards",
+    "truncations",
+    "terminations",
+}
```
pseudo-rnd-thoughts, line 20: Why are we not keeping the info?

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+
+
+class EpisodeMetadataCallback:
+    """Callback to full episode after saving to hdf5 file as a group.
+
+    This callback can be overridden to add extra metada attributes or statistics to
```
pseudo-rnd-thoughts, line 26: `metadata`

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+    each HDF5 episode group in the Minari dataset. The custom callback can then be
+    passed to the DataCollectorV0 wrapper in the `episode_metadata_callback` argument.
```
pseudo-rnd-thoughts, line 28: wrapper **to** the

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+
+    TODO: add more default statistics to episode datasets
```
pseudo-rnd-thoughts, line 30: When will this be done?

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+    Minari).
+
+    This callback can be overridden to add extra environment information in each step or
+    edit the observation, action, reward, termination, truncation, or info returns.
+    """
+
+    def __init__(self, env: gym.Env):
+        self.env = env
+
+        def check_flatten_space(space: gym.spaces.Space):
```
pseudo-rnd-thoughts, line 65: Can you use `space. is_np_flattenable `? https://gymnasium.farama.org/api/spaces/#gymnasium.spaces.space.Space.is_np_flattenable

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+        if not os.path.exists(self.datasets_path):
+            os.makedirs(self.datasets_path)
+
+        self._tmp_dir = tempfile.TemporaryDirectory(dir=self.datasets_path)
+        self._tmp_f = h5py.File(
+            os.path.join(self._tmp_dir.name, "tmp_dataset.hdf5"), "a", track_order=True
+        )  # track insertion order of groups ('episodes_i')
+        self._tmp_f.attrs[
+            "env_spec"
+        ] = self.env.spec.to_json()  # pyright: ignore [reportOptionalMemberAccess]
```
pseudo-rnd-thoughts, line 258: Just `assert self.env.spec is not None` to remove the pyright comment

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+                if isinstance(value, dict):
+                    episode_buffer[key] = self._add_to_episode_buffer(
+                        episode_buffer[key], value
+                    )
+                else:
+                    episode_buffer[key].append(value)
+
+        return episode_buffer
+
+    def step(self, action: Any):
```
pseudo-rnd-thoughts, line 305: Add full type hint

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+        """
+        if action is not None:
+            # Flatten the actions
+            if self.flatten_action:
+                action = spaces.utils.flatten(self.env.action_space, action)
+        # Flatten the observations
+        if self.flatten_observation:
+            obs = spaces.utils.flatten(self.env.observation_space, obs)
+
+        step_data = {
```
pseudo-rnd-thoughts, line 145: I would create a `TypedDict` https://docs.python.org/3/library/typing.html#typing.TypedDict for the step data. 
This will improve your type hint as instead of `Dict` you can have `StepData` which is more descriptive to users

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+            info=info,
+            action=action,
+            rew=rew,
+            terminated=terminated,
+            truncated=truncated,
+        )
+
+        # force step data dicitonary to include keys corresponding to Gymnasium step returns:
+        # actions, observations, rewards, terminations, truncatins, and infos
+        assert STEP_DATA_KEYS.issubset(step_data.keys())
```
pseudo-rnd-thoughts, line 322: `assert isinstance(step_data, StepData)`

minari/utils/data_collector.py
```diff
@@ -0,0 +1,573 @@
...
+            # New episode
+            self._episode_id += 1
+
+        # add new episode buffer to global buffer when episode finishes with truncation or termination
+        if clear_buffers or step_data["terminations"] or step_data["truncations"]:
+            self._buffer.append({key: [] for key in STEP_DATA_KEYS})
+
+        return obs, rew, terminated, truncated, info
+
+    def reset(self, seed: Optional[int] = None, *args, **kwargs):
```
pseudo-rnd-thoughts, line 366: Fix the type hints and remove args / kwargs to options

pseudo-rnd-thoughts (changes requested):

minari/utils/data_collector.py
rodrigodelazcano: I think `issubset` works better in this case since `step_data` can have extra keys apart from the `STEP_DATA_KEYS`. `STEP_DATA_KEYS` are mandatory keys that have to be returned every step.

minari/utils/data_collector.py
rodrigodelazcano: I can't because I need to check specifically for `Dict` and `Tuple` spaces. The rest of the spaces shouldn't be flattened or are not yet compatible with Minari

minari/utils/data_collector.py
rodrigodelazcano: Haven't set a timeline for it. I have to investigate a bit more what would be useful, open to suggestions. This will probably be done in separate PR.

minari/utils/data_collector.py
rodrigodelazcano: I set recording infos as optional since some users may not find useful what is being returned. The argument is `record_infos` in `DataCollectorV0`, the default is `True` which does record the infos.

pyproject.toml
rodrigodelazcano: I changed it here and improved some of the type hint. thanks!

pseudo-rnd-thoughts (approved):
I think this is good to merge now. 
We need to add more testing in the future but we should release this for people to start using and testing with

PR #54, opened 2023-04-06
Added iterate_episodes and made dataset an iterable

Howuhh:
# Description

Although the documentation says that Minari doesn't serve the purpose of creating replay buffers, the dataset itself only allows you to sample random episodes, which is not very useful in practice for creating custom replay buffers or dataloaders. It's much more useful to be able to iterate or get a generator by episodes (which won't load everything into memory). I am currently using Minari to build dataset for my tasks and this kind of functionality is very lacking for convenient use. So I added the ability to iterate through the episodes. 

## Type of change

# Checklist:


rodrigodelazcano:
Hey @Howuhh thanks a lot for this PR! I'm wondering if merging this new method `iterate_episodes` and `sample_episodes` will be possible. Have a single generator `iterate_episodes`, with a `batch_size` and `shuffle` parameters. Let me know your thoughts about this.

I can merge this PR first after the pre-commit is fixed and make an issue for the single generator implementation

rodrigodelazcano:
Second thoughts, I don't think `batch_size` is necessary since what we are trying to do here is populate a buffer.

Howuhh:
@rodrigodelazcano To be honest, I don't really understand the current issue with the pre-commit, and what's more, locally, all the checks go through without errors (with setup described in `CONTRIBUTING.md`).

minari/dataset/minari_dataset.py
```diff
@@ -218,6 +218,24 @@ def sample_episodes(self, n_episodes: int) -> Iterable[EpisodeData]:
...
+        Args:
+            episode_indices (Optional[Union[np.ndarray, List[int]]], optional): episode indices to iterate over.
+        """
+        if episode_indices is None:
+            episode_indices = self._episode_indices
+
+        if isinstance(episode_indices, np.ndarray) and episode_indices.ndim != 1:
+            raise RuntimeError("The indices must have a ndim of 1.")
+
+        for episode_index in sorted(list(episode_indices)):
```
younik, line 235: Hey, thanks for the PR!

Any reason to sort the indices?

rodrigodelazcano:
> @rodrigodelazcano To be honest, I don't really understand the current issue with the pre-commit, and what's more, locally, all the checks go through without errors (with setup described in `CONTRIBUTING.md`).

Interesting, let me have a look at it.

minari/dataset/minari_dataset.py
Howuhh: Hi @younik! Just that the order of indices is always deterministic. I thought this would save the user from having to think in what order the indexes of the episodes would be passed to iterate.

younik: I would avoid sorting because if I define the argument `episode_indices` in a specific order, I expect the output to match it.

rodrigodelazcano: I agree with @younik. @Howuhh can you remove the sorting?

Howuhh: @rodrigodelazcano Yeah, it would be better without sorting

rodrigodelazcano:
@Howuhh regarding the pre-commit errors I'm also unable to replicate them locally. Looking at the logs it seems that the issue comes from pyright identifying the ndarray type as Unknown. I also think that we can omit the use of a numpy array to pass the list of indices, passing a list should be enough. Can you remove this type and see if the error is fixed?

Howuhh:
@rodrigodelazcano Yup, but then it will be a little bit inconsistent with type of `episode_indices` for `MinariDataset` class.

minari/dataset/minari_dataset.py
Howuhh: Removed sorted

Howuhh:
also, there are examples such as `dataset.name` the documentation, however .name attr does not exist in the code (as far as I can tell)

@rodrigodelazcano I actually don't think it's necessary to remove ]sample_episodes], maybe it could be useful to someone else.

minari/dataset/minari_dataset.py
```diff
@@ -218,6 +218,25 @@ def sample_episodes(self, n_episodes: int) -> Iterable[EpisodeData]:
...
+            episode_indices (Optional[List[int]], optional): episode indices to iterate over.
+        """
+        if episode_indices is None:
+            assert self._episode_indices is not None
+            assert self._episode_indices.ndim == 1
+            episode_indices = self._episode_indices.tolist()
+
+        for (
+            episode_index
+        ) in episode_indices:  # pyright: ignore [reportOptionalIterable]
```
younik, line 236: Is it possible to avoid `#pyright: ignore` here and thus make it one line?

It seems it doesn't recognize in line 140 that episode_indices can't be None, but it should work with an assert

younik (approved):
Looks good to me; I just added a small comment

rodrigodelazcano (approved):
LGTM! I'll merge after @younik reviews are fixed

minari/dataset/minari_dataset.py
rodrigodelazcano: @Howuhh can you edit this?

Howuhh: @rodrigodelazcano yup, been a little busy, I'll correct it

Howuhh: done, check are all green locally, so mb this will work

PR #55, opened 2023-04-07
Add ReplayBuffer

younik:
Adding a draft for ReplayBuffer

### DIFF
--- a/minari/replay_buffer.py
+++ b/minari/replay_buffer.py
@@ -1,0 +1,89 @@
+from typing import NamedTuple
+import numpy as np
+from minari.dataset.minari_dataset import MinariDataset
+import gymnasium as gym
+
+
+class ReplayBuffer:
+
+    def __init__(self, buffer_size: int) -> None:
+        self.max_size = buffer_size
+        self.size = 0
+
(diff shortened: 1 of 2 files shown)


minari/replay_buffer.py
```diff
@@ -0,0 +1,89 @@
...
+            terminations=terminations,
+            truncations=truncations,
+            next_obs=next_obs,
+        )
+
+    def _allocate_memory(self, observation_space: gym.Space, action_space: gym.Space):
+        assert observation_space.shape is not None
+        assert action_space.shape is not None
+    
+        self._obs = np.zeros((self.max_size, *observation_space.shape), dtype=observation_space.dtype)
```
Howuhh, line 69: I think in more general implementation, buffer should use gym.spaces.unflatten to provide observations samples in their original shape

younik: @Howuhh Thanks for the review; I am working on it!

Are you in our Discord channel? Can you drop me a message? (younik)

rodrigodelazcano: Can we make the unflatten optional?

Issue #56, opened 2023-04-07
[Bug Report] h5py error when combining more than 2 datasets


Howuhh:
**Describe the bug**
Have you tried combining more than two dataset at a time? When I try to do that, I get an error, while combining only two datasets works fine. I think this might be related to the https://github.com/h5py/h5py/issues/1385.

**Error**
```shell
Traceback (most recent call last):
  File "rllib_mini/train_ppo.py", line 171, in <module>
    train()
  File "/usr/local/lib/python3.8/dist-packages/pyrallis/argparsing.py", line 158, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "rllib_mini/train_ppo.py", line 164, in train
    comdined_dataset = minari.combine_datasets(datasets, new_dataset_id=f"{config.env_name}-dataset-v{config.version}")
  File "/usr/local/lib/python3.8/dist-packages/minari/utils.py", line 132, in combine_datasets
    combined_data_file.attrs["author"] = dataset_file.attrs["author"]
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/usr/local/lib/python3.8/dist-packages/h5py/_hl/attrs.py", line 103, in __setitem__
    self.create(name, data=value)
  File "/usr/local/lib/python3.8/dist-packages/h5py/_hl/attrs.py", line 212, in create
    h5a.delete(self._id, self._e(name))
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5a.pyx", line 145, in h5py.h5a.delete
KeyError: 'Unable to delete attribute (record is not in B-tree)'
```

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo (**required**)

rodrigodelazcano:
I had this issue in the past, the hdf5 file didn't like having its attributes override. A patch is to instead use the `modify` method.
I already fixed it for the `total_steps` and `total_episodes` attributes, but missed doing it for `email` and `author_email`.
https://github.com/Farama-Foundation/Minari/blob/45a1b6d9327f899e097c944b9086e847112d54c8/minari/utils.py#L125

[...]

PR #82, opened 2023-05-28
Added option to copy data when combining datasets

Howuhh:
# Description

Related to https://github.com/Farama-Foundation/Minari/issues/56 

## Type of change

# Checklist:

### DIFF
--- a/minari/utils.py
+++ b/minari/utils.py
@@ -141,8 +141,8 @@ def combine_datasets(
                 combined_data_file.attrs["total_steps"] + dataset.spec.total_steps,
             )
 
+            # TODO: list of authors, and emails
             with h5py.File(dataset.spec.data_path, "r") as dataset_file:
-                # TODO: list of authors, and emails
                 combined_data_file.attrs.modify("author", dataset_file.attrs["author"])
                 combined_data_file.attrs.modify(
                     "author_email", dataset_file.attrs["author_email"]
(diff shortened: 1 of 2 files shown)


Howuhh:
a bit uglier than initially, but unfortunately because of the limitations of h5py you can't do it like in the first commit

balisujohn (changes requested):
Overall, it looks good, and it seems to perform as intended, creating combined datasets that are not broken when the source datasets are deleted.

I think also it would be good to include a test that makes sure the true-copied combined dataset is still intact after deleting the source datasets.

In a future PR, I think the true copy behavior should be made the default for the `combine_datasets()` function, and also the default for the `combine` command in the cli, though it should still be possible to enable the external link combination behavior in the cli with the optional flag` --external-link` I think we should make there be a check to see if combined datasets depend on other datasets, because right now you can break combined datasets with no warning by deleting their dependency datasets using `minari delete`. 

Curious to hear your thoughts on this.

balisujohn (approved):

balisujohn:
LGTM :^)

balisujohn:
@Howuhh Thanks for your thoughts on this.  I can understand the argument against switching to true copy as the primary method, so I'll hold off on making that change for the time being. I will make a task to add am optional  command line flag to enable true copy to the cli command `minari combine`.

PR #86, opened 2023-05-30
Checking env base lib installation

shreyansjainn:
# Description

Addressing the `ImportErrror` occurring while downloading the dataset and if the base library for the environment is not installed.

## Type of change

# Checklist:


minari/dataset/minari_storage.py
```diff
@@ -43,7 +45,15 @@ def __init__(self, data_path: PathLike):
...
 
-            env = gym.make(self._env_spec)
+            # checking if the base library of the environment is present in the environment
+            entry_point = json.loads(f.attrs["env_spec"])["entry_point"]
+            base_lib = entry_point.split(".")[0]
+            env_name = self._env_spec.id
+
+            if base_lib in sys.modules:
+                env = gym.make(self._env_spec)
+            else:
```
elliottower, line 55: couldn't this just be done with a simple try-except? in other farama projects we've just explicitly caught `ImportError`s, that way you have the full error message from python (easier debugging/more consistent for user to see) as well as our custom text like you added (mentioning the environment name, etc)

balisujohn: Assuming we want to raise a new error in the catch block, we may need to add a `DependencyNotInstalled` error to Minari, or use Gymnasium's `DependencyNotInstalled` error. @elliottower what are your thoughts on this?

shreyansjainn: @elliottower my initial thought was to go ahead with try-except only, but i didn't see any try-except block in the whole codebase and saw lot of if conditions instead, so that is why i went with this route. i can add try-block too if that is acceptable. and like @balisujohn mentioned, if we can add `DependencyNotInstalled` error, that will be a much clearer error i feel compared to `importError`

balisujohn: ok sounds good, I think then it would be best to add a `DependencyNotInstalled` error in `error.py` mirroring the way it's done in Gymnasium, then throw that in the catch block for the `importError`

shreyansjainn: one question, for that to be properly used in the minari repo, we will have to wait till the next gymnasium release right?

balisujohn: I meant to create a new `DependencyNotInstalled` error class in Minari, just copying the way it's done in Gymnasium, so it shouldn't depend on Gymnasium at all.

balisujohn (approved):
LGTM :^)

Issue #79, opened 2023-05-26
Code review notes


RedTachyon:
Writing this without fixes for now, things I'm noticing going through the library:

1. The dataset listed in README doesn't seem to exist. This is bad, as that's the first thing a beginner will likely try when trying to interact with the library. We should either swap it, or create it.
2. Remote datasets are `'door-cloned-v0', 'door-expert-v0', 'door-human-v0', ...`, at least reading it initially. One issue here is that the dataset names don't really tell you much unless you already know the context - we should probably mention that they're specifically for mujoco, probably specifically for some specific model - maybe we can list the name of the gymnasium env?
3. Trying to do `minari.download_dataset("door-human-v0")` results in a pretty ugly/uncaught import error for `gymnasium_robotics`. Since gymnasium-robotics isn't a necessary dependency, we should have a more elegant way of handling this.
4. (3a) The import issue seems to happen because when downloading the dataset, we're creating the environment to get the obs and action spaces. Ignoring even the dependency issue, this seems very redundant - can't we just save the obs/action spaces in the dataset? They should be easy to serialize.
5. Downloading the same dataset twice repeats the download and overwrites it. We explicitly acknowledge this, but imo there should be an `overwrite` or `force` argument to do that behavior, and by default it should just skip downloading it. It might be worthwhile to think if we want to do explicit downloads like this in the first place, but that's another thing.
6. `EpisodeData` is a namedtuple, which leads to the default representation being an enormous printout of all observations, rewards etc. This is not very useful, so we should probably make it a proper dataclass instead, and override the default representation.

Stopping here for now, @rodrigodelazcano let me know if there's anything I'm missing anything for any of those points (like some additional reasons for doing X thing in a certain way that I'm unaware of). Otherwise we can make this into a to-do list of improvements

balisujohn:
Checklist for completing/addressing code review.

- [ ] 1 https://github.com/Farama-Foundation/Minari/pull/80

- [ ] 2

- [ ] 3

- [ ] 4

- [ ] 5

- [ ] 6

shreyansjainn:
Hi @balisujohn, a couple of pointers mentioned in this issue I wanted to get feedback on. 

- for the 2nd one, like Ariel suggested, I think we can name the datasets starting with the env name instead of just mentioning the shorthand. something like AdroitHandDoor-human-v0 instead of just door-human-v0. here I've not mentioned the version of the env as it can cause some confusion having two versions mentioned in the name.

[...]

PR #88, opened 2023-06-08
ENH: EpisodeData as dataclass

younik:
# Description

Make `EpisodeData` a dataclass instead of `NamedTuple` and implement `__repr__` method as suggested in https://github.com/Farama-Foundation/Minari/issues/79

## Type of change

# Checklist:

### DIFF
--- a/tests/dataset/test_minari_dataset.py
+++ b/tests/dataset/test_minari_dataset.py
@@ -1,3 +1,5 @@
+import re
+
 import gymnasium as gym
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


balisujohn (commented):
I think it would be good to add a test for this, but otherwise it looks good to me.

younik:
> I think it would be good to add a test for this, but otherwise it looks good to me.

I added a test for it, thanks!

For testing, I copied the list of spaces that you used in your PR and added them to a file `test_common`; if we merge this first, you should update your `test_serialization` to use it

balisujohn:
Thanks for the test. I like the shared dependency for the test spaces so we avoid repeating definitions, I'll copy this pattern for the Dummy space registration for the unflattened spaces PR. 

For the new test, can you also add an assert to the test that checks to see that the `repr` matches the expected `repr` string for each space?

younik:
> Thanks for the test. I like the shared dependency for the test spaces so we avoid repeating definitions, I'll copy this pattern for the Dummy space registration for the unflattened spaces PR.
> 
> For the new test, can you also add an assert to the test that checks to see that the `repr` matches the expected `repr` string for each space?

done

balisujohn (approved):
LGTM :^)

PR #77, opened 2023-05-24
[Dataset Format Standard Modification] Record Unflattened Spaces in Datasets

rodrigodelazcano:
# Description

This PR is the continuation of issue #57. The PR removes the automatic flattening of observation/action spaces in `StepDataCallback` to store `Dict` and `Tuple` spaces in their original format. This will allow to create custom dataset spaces different from the environment.

## Next tasks
- [x] Find solution to record `Tuple` spaces (maybe name each hdf5 group/dataset as `element_{id}`)
- [ ] Encode custom action/observation spaces for datasets (the spaces can be different from the environment's)
- [x] Fix pytest
- [ ] Documentation update 

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have run `pytest -v` and no errors are present.
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #91, opened 2023-06-13
[WIP] Datacollector refactoring

RedTachyon:
I'm preparing to make a vectorized data collector, but along the way I started disentangling some of the code. Right now, I think the changes only consist of extracting a few methods which might as well be separate functions. It shouldn't cause any changes in actual behavior.

Please let me know if this makes sense to do or if I'm missing something internal about Minari, in the meantime I'll keep going to make the codebase structured a little bit better, and then hopefully reuse those components for the vectorized version.


younik:
This makes a lot of sense; `DataCollector` urges a refactor. I would wait for big changes tho, as [this PR](https://github.com/Farama-Foundation/Minari/pull/77) is going to be merged soon

minari/data_collector/data_collector.py
```diff
@@ -461,3 +379,105 @@ def close(self):
         # Close tmp_dataset.hdf5
         self._tmp_f.close()
         shutil.rmtree(self._tmp_dir.name)
+
+
+def clear_buffer(
```
younik, line 384: Can this and`add_to_episode_buffer`become a method of `EpisodeBuffer` (after making it a proper class)?

alexdavey:
It appears that the current data collector does not copy over the infos from the previous episode when it was terminated without reset. I have fixed this issue [here](https://github.com/Farama-Foundation/Minari/compare/main...alexdavey:Minari:datacollector_infos).

I'm happy to send a separate PR, but it will clash with this refactoring. @balisujohn suggested I reference this so you are aware. From what I can see the issue persists here, and also in the [other PR](https://github.com/Farama-Foundation/Minari/pull/77) in this thread.

PR #94, opened 2023-06-20
HuggingFace Integration

balisujohn:
# Description

Draft PR for hugging face Minari integration. 

Adds functions to convert back and forth between `MinariDataset` and `datasets.Dataset` from Hugging Face datasets. Additionally, it adds functions that allow the user to push and pull datasets from hugging face hub. The core code is ready for review, but there are still a few more features I will add, for which I'm adding a checklist to this description:

I also refactored the tests slightly, creating the new helperful function `create_dummy_dataset_with_collecter_env_helper` to avoid code repetition.

# Additional Features
- [ ] CLI wrappers for push and pull dataset.
- [ ] Wrappers to to automatically convert minari datasets when uploading and downloading.
- [ ] Support for Text Spaces

# Checklist:


younik (changes requested):
I added few comments on the functions, but overall I have some concerns on the API.

For the user, the current workflow is something like that:

```python
import gymnasium as gym
import minari
from minor import DataCollectorV0
from minari.integrations.hugging_face import (
    convert_hugging_face_dataset_to_minari_dataset,
    convert_minari_dataset_to_hugging_face_dataset,
    pull_dataset_from_hugging_face,
    push_dataset_to_hugging_face,
)

env = DataCollectorV0(gym.make("EnvName"))
... # code that generates the dataset
dataset = minari.create_dataset_from_collector_env(...)

hf_dataset = convert_minari_dataset_to_hugging_face_dataset(dataset)
push_dataset_to_hugging_face(hf_dataset, "name/repo")
```
and then
```python
hf_dataset = pull_dataset_from_hugging_face("name/repo")
dataset = convert_hugging_face_dataset_to_minari_dataset(hf_dataset)
```

The main red flag for me here is that we have public functions that are specifically for huggingface. I think this should be (more) transparent to the user, i.e. the function in `minari/integrations/hugging_face.py` are all private.

I have a couple of alternatives in mind:

### Add a HF flag to load and upload
```python
dataset = minari.load_dataset("name/repo", hugging_face_hub=True)
```
Which pull the dataset and return a `MinariDataset` that reads from HF dataset using the HuggingFaceStorage.

Similarly for pushing:
```python
minari.upload_dataset('dataset-name', hugging_face_hub=True)
```
Cons of this: 
 - we still have flags specifically for HF.
  - Load dataset directly from cloud, while we have a function `download_dataset`. I imagine, in the future, we also want the possibility to stream directly from the cloud: shall we drop `download_dataset`, and implicitly download on `load_dataset`?
  
  ### Use a setup_remote()
We discussed about having the possibility to setup different remotes than our GCP bucket. This is a particular case of that. We can create the API for that and use it in this case:

```python
minari.setup_remote(
      "https://huggingface.co/balisujohn",
      # others args like auth_key
)
```

And now every `load_dataset`/`upload_dataset` takes from/push to the HF hub directly as before, and conversions are done under the hood.
Cons:
 - Still, `download_dataset` it is no-sense
 - `setup_remote` should work also for GCP bucket and who knows
 - It complicates the library code as `setup_remote` changes other function behavior
 
 
 I am more prone for the second version, but it also requires more work

minari/integrations/hugging_face.py
```diff
@@ -0,0 +1,212 @@
...
+            f"_index_{str(index)}": _reconstuct_obs_or_action_at_index_recursive(
+                entry, index
+            )
+            for index, entry in enumerate(data)
+        }
+
+    elif isinstance(data, np.ndarray):
+        return data[index]
+    else:
+        assert False, f"error, invalid observation or action structure{data}"
```
RedTachyon, line 36: Is there a reason to `asset False` instead of just raising an exception?

minari/integrations/hugging_face.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Create Minari dataset and store locally
+    dataset = minari.create_dataset_from_buffers(
+        dataset_id=dataset_id,
+        env=gym.make(env_name),
+        buffer=buffer,
+        algorithm_name="random_policy",
+        code_permalink="https://github.com/Farama-Foundation/Minari/blob/f095bfe07f8dc6642082599e07779ec1dd9b2667/tutorials/LocalStorage/local_storage.py",
+        author="WillDudley",
+        author_email="wdudley@farama.org",
```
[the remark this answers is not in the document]
RedTachyon: Probably settable from function arguments?

RedTachyon (commented):
Note: I didn't do a comprehensive review of the code, just left a few comments for whatever stood out to me.

As for the higher-level design, I don't really mind having public functions in a separate `minari.integrations.hugging_face` namespace, if those are the ones that developers would use for pulling/uploading/converting. 
I'd potentially consider making the conversion back and forth automatic inside of the push/pull functions, depending on whether or not it makes any sense to operate on "raw" HF datasets in the context of Minari.

@younik can you elaborate on your issue with those functions being public? Imo the namespace makes it explicit enough, but I might be missing something.


As for the two alternative proposals:
- I'm not necessarily a fan of integrating it into the core `load_dataset` etc functions, we'd essentially tie core functionality to an external library and external servers. Keeping integrations separate (but accessible) is the right move imo
- `setup_remote` sounds like a somewhat more ambitious plan for the future, like being generic between GCP/AWS/Azure/HF/whatever else, so my guess is that it's not a solution for right now?

younik:
@RedTachyon 
One thing that it is likely to happen is that we support the HF data format, and then the conversion functions are no-op. We have `MinariStorage` that is designed to abstract to the dataset the difference on file formats (see https://github.com/Farama-Foundation/Minari/pull/94#discussion_r1241855457)

Also, if we want to support other libraries as well (e.g. [RLDS](https://github.com/google-research/rlds)), we need other conversion functions; I don't think this is interesting for the user.

And we may want to add some loading keywords to `load_dataset` and then they must be added also to `convert_hugging_face_dataset_to_minari_dataset` with the same semantics.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "numpy >=1.21.0",
     "h5py>=3.8.0",
     "tqdm>=4.65.0",
+    "datasets>=2.13.0",
```
elliottower, line 31: Should this be an optional requirement? Like install minari[huggingface]

PR #96, opened 2023-06-20
Copy infos from previous episode

alexdavey:
# Description

Currently, the infos step data from the previous time step is not copied if the environment was truncated/terminated without reset. This PR fixes that. The number of infos per episode is now always the same as the number of observations.

This was mentioned previously in issue #91.

## Type of change

# Checklist:


minari/data_collector/data_collector.py
```diff
@@ -209,13 +209,19 @@ def step(
...
+                    {}, self._previous_eps_final_info
+                )
             self._new_episode = False
 
         # add step data to last episode buffer
         self._buffer[-1] = self._add_to_episode_buffer(self._buffer[-1], step_data)
 
         if step_data["terminations"] or step_data["truncations"]:
             self._previous_eps_final_obs = step_data["observations"]
+            if self._record_infos:
```
rodrigodelazcano, line 223: Can you add a small comment on why we are saving the lat obs and info?

rodrigodelazcano (commented):
LGTM, just the tests missing :). @balisujohn any suggestions?

minari/data_collector/data_collector.py
rodrigodelazcano, line 223: Can you add a small comment on why we are saving the last obs and info?

rodrigodelazcano (approved):

minari/data_collector/data_collector.py
alexdavey: Done. Also updated the comment above to reflect the fact that infos are also being saved.

balisujohn:
LGTM :^) Skipping adding the test for now, because we need to add somewhat involved tests for info.

PR #99, opened 2023-06-23
ENH: Support text spaces

younik:
# Description

Add support for `gym.spaces.Text`

## Type of change

# Checklist:


minari/data_collector/data_collector.py
```diff
@@ -348,76 +348,60 @@ def clear_buffer(dictionary_buffer: EpisodeBuffer, episode_group: h5py.Group):
...
-                            current_dataset_shape = episode_group[key].shape[0]
-                        else:
-                            current_dataset_shape = self._last_episode_n_steps
-                            if key == "observations":
-                                current_dataset_shape += (
-                                    1  # include initial observation
-                                )
+                        current_dataset_shape = episode_group[key].shape[0]
+                        if key == "observations":
+                            current_dataset_shape += 1
```
younik, line 393: @rodrigodelazcano I tried refactoring this else branch

Important changes:
 - the length of the episode is always taken from `episode_group[key].shape[0]` without using `self._last_episode_n_steps`. Shouldn't they always be the same?
- Always specify the `maxshape` argue ven when you dump everything (not mutable). It allows to merge the two branches
- Add `dtype` specification on create_dataset. It uses the dtype of the ndarray

younik:
For now it only supports `UTF-8`; it should be possible to support `unicode` using NumPy object arrays; it is just less memory efficient

minari/data_collector/data_collector.py
```diff
@@ -348,76 +348,60 @@ def clear_buffer(dictionary_buffer: EpisodeBuffer, episode_group: h5py.Group):
...
                     clear_buffer(dict_data, eps_group_to_clear)
                 else:
-                    # convert data to numpy
-                    np_data = np.asarray(data)
-                    assert np.all(
-                        np.logical_not(np.isnan(np_data))
-                    ), "Nan found after cast to nump array, check the type of 'data'."
+                    if all(map(lambda elem: isinstance(elem, str), data)):
+                        data_shape = (len(data),)
+                        dtype = h5py.string_dtype(encoding="utf-8")
```
elliottower, line 378: Oh I see how this is a bottleneck as there's only this option or ascii (which is less chars), so given that I think it's fine to do utf-8 https://docs.h5py.org/en/stable/special.html#h5py.string_dtype

elliottower: Did some more searching around and this shouldn't be a problem after all, you can still encode full unicode characters using utf-8 (on unix they are encoded that way by default, on windows apparently it's utf-16), it just uses multi-byte encoding (https://unicodebook.readthedocs.io/unicode_encodings.html). Not sure how exactly it would work with gymnasium spaces but I think it should be good enough, utf-8 is the standard for pretty much everything. If a user really needs to use unicode characters in their gymnasium action/obs space I think they should be able to set the charset as just utf-8 and encode the unicode characters into utf-8.

tests/common.py
```diff
@@ -143,6 +143,23 @@ def reset(self, seed=None, options=None):
         return self.observation_space.sample(), {}
 
 
+class DummyTextEnv(gym.Env):
+    def __init__(self):
+        self.action_space = spaces.Text(max_length=10, min_length=2, charset="01")
+
+        self.observation_space = spaces.Text(max_length=20)
```
elliottower, line 150: You should probably set `Text(max_length=20, charset=string.printable)` as that will be a list of all the printable ascii chars (length 100) rather than just the letters a-z and digits which is the default for gymnasium. You could have the charset be all of the utf-8 characters but that would be a million (as all unicode chars can be encoded) so sampling from that will probably take a while and not be ideal.
If you want the printable unicode multilingual characters (65,000) length it would be `utf8_characters = [chr(i) for i in range(0x10000) if unicodedata.category(chr(i)) != "Cc"]`, but that's probably not necessary besides maybe showing people that it can be done.

balisujohn (changes requested):
I think there should be a test which uses full UTF-8 as the charset for a Text Space, similarly to how @elliottower described for the unicode multilingual characters. https://github.com/Farama-Foundation/Minari/pull/99/files#r1240189639

For this, I'm curious if, we might need a special flag to mean full utf-8 when serializing, instead of serializing a 1mil+ character string.


I'm leaning towards asking for a way to test also that the strings which get saved are exactly the same as the strings you get when loading the episode and not just check that they are in the space like we are currently doing. We could improve the dataset creation tests generally to enforce this for all space types, but let's say this last paragraph is a suggestion for a later PR.

elliottower:
> I think there should be a test which uses full UTF-8 as the charset for a Text Space, similarly to how @elliottower described for the unicode multilingual characters. https://github.com/Farama-Foundation/Minari/pull/99/files#r1240189639
> 
> For this, I'm curious if, we might need a special flag to mean full utf-8 when serializing, instead of serializing a 1mil+ character string.
> 
> I'm leaning towards asking for a way to test also that the strings which get saved are exactly the same as the strings you get when loading the episode and not just check that they are in the space like we are currently doing. We could improve the dataset creation tests generally to enforce this for all space types, but let's say this last paragraph is a suggestion for a later PR.

More I think about it the more feel like the gymnasium spaces are kind of irrelevant to serializing, no matter what it’ll always go into the memory as utf8 encoding, people can do whatever subset of Unicode that they want and it’ll always be encodable that way.

balisujohn:
In general, I think that providing space serialization means a contract that the user will be able to fully recover the space specification beyond what is implicitly present in the data itself, since we provide bounds on box and discrete spaces, and I think we need to provide similar metadata on serialized text spaces. Even if we don't need to know the space to faithfully reproduce the data, the user still needs to be able to recover the space from the serialization. I don't think it will add too much complexity if we add a boolean flag to a `Text` space serialized json that is set to true if all unicode characters are supported and false otherwise, and if it is set to true, the charset is left empty in the serialization. 

We might even want to consider upstreaming something similar into `Gymnasium.spaces.Text` (because with the actual Text spaces in  mempry, we will stilll have to use the million plus character long charset string, which could be represented as a couple bytes with either boundary-based charsets or a special "all utf-8" flag.

elliottower:
> In general, I think that providing space serialization means a contract that the user will be able to fully recover the space specification beyond what is implicitly present in the data itself. We provide bounds on box and discrete spaces, and I think we need to provide similar metadata on serialized text spaces. Even if we don't need to know the space to faithfully reproduce the data, the user still needs to be able to recover the space from the serialization. I don't think it will add too much complexity if we add a boolean flag to a `Text` space serialized json that is set to true if all unicode characters are supported and false otherwise, and if it is set to true, the charset is left empty in the serialization.
> 
> We might even want to consider upstreaming something similar into `Gymnasium.spaces.Text` (because with the actual Text spaces in mempry, we will stilll have to use the million plus character long charset string, which could be represented as a couple bytes with either boundary-based charsets or a special "all utf-8" flag.

Fair enough about users needing to know. We should talk to Mark about it, I think a reasonable thing for gymnasium spaces would be to have encoding options like ascii or utf-8, but we should probably look into things more and get a better idea of what users will want. Last we talked Mark and Ariel said it probably didn’t make sense to have the charset argument or at least there should be easy ways to do utf-8 without specifying a list of every valid character

younik:
For how `Text` space is right now, I don't think it makes sense to have a special case for the full UTF-8 charset. As the user needs to specify it and all the characters are saved in a variable, I think it makes sense to save all the characters as well.
If in Gymnasium they change it (for example, they make the full UTF-8 set as the default value), then it makes sense to support it better here. In that case, a boolean flag or simply not saving the field `charset` (and using the UTF-8 charset as default while loading) makes more sense.

tests/common.py
younik: Trying with all unicode it fails with an informative error `UnicodeEncodeError: 'utf-8' codec can't encode character '\udb25' in position 2: surrogates not allowed`

minari/data_collector/data_collector.py
rodrigodelazcano: 1. Yes this is better. They should always be the same. The initial intention of `self._last_episode_n_steps` was to not read this value directly from file. But I agree this doesn't make any sense if we are doing it for other cases so I approve the change. However, with these changes you should remove the condition for when the key is `observations` is not required since the +1 will  be already present in the hdf5 `observations` dataset.
2. Agree
3. Very necessary. Agree

tests/common.py
```diff
@@ -338,7 +364,9 @@ def register_dummy_envs():
                                             gym.spaces.Box(
                                                 low=4, high=5, dtype=np.float32
                                             ),
-                                            gym.spaces.Text(1),
+                                            gym.spaces.Graph(
```
rodrigodelazcano, line 368: Why the `Graph` space? Is Minari already compatible with this type of space?

rodrigodelazcano (changes requested):

tests/common.py
younik: This list is used to test unsupported spaces; thus I removed `Text` and add `Graph`

tests/common.py
balisujohn: We need to somehow create a charset for all legal unicode characters, and decide what our stance is on surrogate characters, this may be nontrivial. Naively, I tried:

````
utf8_charset = "".join([chr(i) for i in range(sys.maxunicode)])
````
and I got a similar error
````
E   UnicodeEncodeError: 'utf-8' codec can't encode character '\ud937' in position 3: surrogates not allowed
````

Here is a  h5py doc page talking about surrogate characters:
https://docs.h5py.org/en/3.2.1/strings.html

Comment (younik) [reply tests/common.py]: I don't understand what you mean with "decide what our stance is on surrogate characters"

File: tests/common.py [hunk 5]
```diff
@@ -435,6 +464,10 @@ def check_data_integrity(data: MinariStorage, episode_indices: Iterable[int]):
             obs = _reconstuct_obs_or_action_at_index_recursive(
                 episode["observations"], i
             )
+            if not data.observation_space.contains(obs):
+                import pdb
+
+                pdb.set_trace()
```
balisujohn, line 470: I think this is leftover from debugging

tests/common.py
```diff
@@ -143,6 +143,23 @@ def reset(self, seed=None, options=None):
         return self.observation_space.sample(), {}
 
 
+class DummyTextEnv(gym.Env):
+    def __init__(self):
+        self.action_space = spaces.Text(max_length=10, min_length=2, charset="01")
```
balisujohn, line 148: I think it would be good to create a dummy env with a full unicode charset minus surrogate chars that returns emoji as observations, as a stress test.

balisujohn (changes requested):

tests/common.py
younik: This is the case for `DummyTextEnv` for observation, isn't it? 
Do you want it also for action space?

balisujohn (commented):
Ready to merge, once the doc is updated to reflect text space support :^)

rodrigodelazcano (approved):

balisujohn (approved):
LGTM :^)

PR #101, opened 2023-06-28
Removes h5py import in minari_dataset.py, adds tests for minari_dataset.py

balisujohn:
# Description

This PR removes the h5py import in `minari_dataset.py` and also adds some tests for minari_dataset.py. 

This PR does make one breaking change to the API and a seperate slight relaxation of allowed types, both are mentioned in comments.

# Checklist:


minari/dataset/minari_dataset.py
```diff
@@ -253,7 +201,7 @@ def filter_episodes(self, condition: Callable[[h5py.Group], bool]) -> MinariData
         ```
 
         Args:
-            condition (Callable[[h5py.Group], bool]): callable that accepts an episode group and returns True if certain condition is met.
```
balisujohn, line 256: This relaxes the type expected as an input for the filter function to be applied to the underlying episode dataset to be any type, since MinariDatset is no longer supposed to assume the types used in `MinariStorage`

minari/dataset/minari_dataset.py
```diff
@@ -43,56 +41,6 @@ def parse_dataset_id(dataset_id: str) -> tuple[str | None, str, int | None]:
     return env_name, dataset_name, version
 
 
-def clear_episode_buffer(episode_buffer: Dict, episode_group: h5py.Group) -> h5py.Group:
```
balisujohn, line 46: Non private function moved to `minari_storage.py` so this is breaking.

minari/dataset/minari_dataset.py
younik: How about EpisodeData?

minari/dataset/minari_dataset.py
```diff
@@ -307,31 +255,16 @@ def update_dataset_from_collector_env(self, collector_env: DataCollectorV0):
...
-                "total_steps", file.attrs["total_steps"] + new_data_total_steps
-            )
+        self._data.update_from_collector_env(
+            collector_env, new_data_file_path, self._additional_data_id
+        )
+
         self._additional_data_id += 1
 
+        self._episode_indices = np.arange(self._data.total_episodes)
+
```
younik, line 265: It breaks filtered `MinariDataset`; you should append the new indices

minari/dataset/minari_dataset.py
```diff
@@ -347,42 +280,11 @@ def update_dataset_from_buffer(self, buffer: List[dict]):
...
-
-                # TODO: save EpisodeMetadataCallback callback in MinariDataset and update new episode group metadata
-
-            file.attrs.modify("total_episodes", last_episode_id + len(buffer))
-            file.attrs.modify(
-                "total_steps", file.attrs["total_steps"] + additional_steps
-            )
+        self._data.update_from_buffer(buffer, self.spec.data_path)
+
+        self._episode_indices = np.arange(self._data.total_episodes)
```
younik, line 285: same as above

tests/utils/test_dataset_combine.py
```diff
@@ -90,6 +90,7 @@ def _generate_dataset_with_collector_env(dataset_id: str, num_episodes: int = 10
 
 
 def test_combine_datasets():
+
```
younik, line 93: probably not intended

younik (changes requested):
Looks good, just a couple of comments

minari/dataset/minari_dataset.py
balisujohn: Right now `MinariStorage.apply()` calls the filter function with `h5py.group` representations of episodes. We'd need to make it apply pass  `EpisodeData` into the filter function instead of `h5py.group` Curious to hear your thoughts on this.

minari/dataset/minari_dataset.py
balisujohn: done

minari/dataset/minari_dataset.py
balisujohn: done

tests/utils/test_dataset_combine.py
balisujohn: fixed

balisujohn:
One comment is that if env.spec.total_episodes is supposed to reflect the total number of episodes in the underlying MinariStorage(without respect to the applied filter), we should either make it a computed property that loads the `self._data.total_episodes` or we should actually move spec to the `self._data` MinarStorage` instance, because right now if you create a filtered dataset that points to the same `MinariStorage` as another one then add episodes to  the filtered one, the spec doesn't get updated on the other one. Curious to hear your thoughts on this @younik

minari/dataset/minari_dataset.py
younik: From the user perspective, it makes much more sense to have `EpisodeData`; however this means that we should convert the `h5py.group` into `EpisodeData` inside the function. 
This means we add a dependency `MinariStorage` -> `EpisodeData`, and now it makes sense to also change the return type of `get_episodes` to be already `EpisodeData`.

The other alternative is to have `dict`; I think this is more weird for the user, because it is not immediately clear what keys you can use. On the other side it gives more flexibility for possible future `MinariStorage`.

A solution in the middle is having `EpisodeData` in `MinariDataset` and `dict` in `MinariStorage`; then you need to wrap the function in `MinariDataset` to convert the `dict` to `EpsiodeData`. It complicates a bit more the code, but it has the advantage of both. 

I am towards the third option, but I don't have a strong opinion on this; feel free to choose one of the three if you see something else.

younik:
> One comment is that if `env.spec.total_episodes` is supposed to reflect the total number of episodes in the underlying `MinariStorage`(without respect to the applied filter), we should either make it a computed property that loads the `self._data.total_episodes` or we should actually move spec to the `self._data` `MinariStorage` instance, because right now if you create a filtered dataset that points to the same `MinariStorage` as another one then add episodes to the filtered one, the spec doesn't get updated on the other one. Curious to hear your thoughts on this @younik

It is hard for me to comment on this, because it is not clear to me the use-case of `spec`. For the little I understand about it, we want to keep in in `MinariDataset` so, if we want `total_episodes` to reflect the unfiltered total number, we should make it a computed property.

balisujohn:
I talked to Rodrigo, and it seems like spec is supposed to reflect the `MinariDataset` and not the underlying `MinariStorage`, so it's ok that adding episodes to one `MinariDataset` that points to a `MinariStorage` doesn't update `total_episodes` in the the `spec` of the other.

minari/dataset/minari_dataset.py
balisujohn: apply actually gives the condition function a h5py group rather than a dict. Do you think we should change apply to give dicts?

balisujohn: I think I understand, I'm tentatively changing apply to give the condition function `dict`s and appropriately wrapping the condition function in `filter_episodes`

minari/dataset/minari_dataset.py
```diff
@@ -220,7 +176,7 @@ def total_steps(self):
         """Total episodes steps in the Minari dataset."""
         if self._total_steps is None:
             t_steps = self._data.apply(
-                lambda episode: episode["total_steps"],
+                lambda episode: episode.total_steps,
                 episode_indices=self._episode_indices,
             )
             self._total_steps = sum(t_steps)
```
younik, line 182: Now that `total_steps`is computed at init, you can delete this function.
This was intended for lazy initialization in case of large dataset

minari/dataset/minari_dataset.py
```diff
@@ -194,12 +140,22 @@ def __init__(
...
         self._episode_indices = episode_indices
 
+        assert self._episode_indices is not None
+
+        total_steps = sum(
+            [
+                episode["total_timesteps"]
+                for episode in self._data.get_episodes(self._episode_indices.tolist())
+            ]
+        )
```
younik, line 153: I think it is better to use `apply` (as done in the property `total_steps`).
This because in future apply may exploit parallelism

minari/dataset/minari_dataset.py
```diff
@@ -255,9 +213,15 @@ def filter_episodes(self, condition: Callable[[h5py.Group], bool]) -> MinariData
         ```
 
         Args:
-            condition (Callable[[h5py.Group], bool]): callable that accepts an episode group and returns True if certain condition is met.
+            condition (Callable[[Any], bool]): callable that accepts any type(For our current backend, an h5py episode group) and returns True if certain condition is met.
```
younik, line 216: Why not `Callable[EpisodeData]`?

And then the comment updated accordingly (it should not mention h5py): `callable that accepts EpisodeData and returns True if certain condition is met` or `boolean function on EpsiodeData`

minari/dataset/minari_dataset.py
```diff
@@ -243,7 +199,9 @@ def set_seed(self, seed: int):
         """Set seed for random episode sampling generator."""
         self._generator = np.random.default_rng(seed)
 
-    def filter_episodes(self, condition: Callable[[h5py.Group], bool]) -> MinariDataset:
+    def filter_episodes(
+        self, condition: Callable[[EpisodeData], bool]
+    ) -> MinariDataset:
         """Filter the dataset episodes with a condition.
 
         The condition must be a callable with  a single argument, the episode HDF5 group.
```
younik, line 207: remove mention to HDF5, now should be EpisodeData

minari/dataset/minari_dataset.py
```diff
@@ -309,31 +273,28 @@ def update_dataset_from_collector_env(self, collector_env: DataCollectorV0):
...
+            self._episode_indices, np.arange(old_total_episodes, new_total_episodes)
+        )  # ~= np.append(self._episode_indices,np.arange(self._data.total_episodes))
+
+        self.spec.total_episodes = len(self._episode_indices)
+        self.spec.total_steps = sum(
+            [
+                episode["total_timesteps"]
+                for episode in self._data.get_episodes(self._episode_indices)
+            ]
+        )
```
younik, line 296: Better to use apply, as above

minari/dataset/minari_dataset.py
```diff
@@ -349,42 +310,23 @@ def update_dataset_from_buffer(self, buffer: List[dict]):
...
+
+        self._data.update_from_buffer(buffer, self.spec.data_path)
+
+        new_total_episodes = self._data._total_episodes
+
+        self._episode_indices = np.append(
+            self._episode_indices, np.arange(old_total_episodes, new_total_episodes)
+        )  # ~= np.append(self._episode_indices,np.arange(self._data.total_episodes))
+
+        self.spec.total_episodes = len(self._episode_indices)
```
younik, line 323: (a whim): be consistent with init (there you used `.size`)

minari/dataset/minari_storage.py
```diff
@@ -85,7 +87,21 @@ def apply(
             for ep_idx in episode_indices:
                 ep_group = file[f"episode_{ep_idx}"]
                 assert isinstance(ep_group, h5py.Group)
-                out.append(function(ep_group))
+                ep_dict = {
```
younik, line 90: on `apply` signature, change the signature of `function` to accept dictionary

younik (changes requested):
minor changes, then looks good

minari/dataset/minari_dataset.py
younik: (and assign the value to `self.total_steps`)

minari/dataset/minari_dataset.py
```diff
@@ -213,7 +207,7 @@ def filter_episodes(
         ```
 
         Args:
-            condition (Callable[[Any], bool]): callable that accepts any type(For our current backend, an h5py episode group) and returns True if certain condition is met.
+            condition (Callable[[EpisodeData], bool]): callable that accepts any type(For our current backend, an h5py episode group) and returns True if certain condition is met.
```
younik, line 210: Update `any type` with EpisodeData and remove parenthesis `(For our current backend, an h5py episode group)`

minari/dataset/minari_dataset.py
```diff
@@ -162,24 +162,18 @@ def __init__(
...
     @property
     def total_steps(self):
         """Total episodes steps in the Minari dataset."""
-        if self._total_steps is None:
-            t_steps = self._data.apply(
-                lambda episode: episode.total_steps,
-                episode_indices=self._episode_indices,
-            )
-            self._total_steps = sum(t_steps)
         return self._total_steps
```
younik: I would totally remove the function and change the name of the attribute `_total_steps` to `total_steps` 
[It is not pythonic to have private attribute with getter without a good reason]

younik (approved):
two small comments and ready to merge for me

minari/dataset/minari_dataset.py
balisujohn: done

minari/dataset/minari_dataset.py
balisujohn: done

minari/dataset/minari_dataset.py
balisujohn: done

minari/dataset/minari_dataset.py
balisujohn: done

minari/dataset/minari_dataset.py
balisujohn: done

minari/dataset/minari_storage.py
balisujohn: done

PR #103, opened 2023-06-30
Completed the assigned task and fixed one bug 

Aaisha-Rani:
* Added gymnasium Env id, File size on disk, and dataset group column to the Minari list local cli commands 

* Added dataset_group attribute, with  string Unknown as a placeholder value.

* Updated the Docs Data set standard and individual dataset pages.

* Fixed a bug  in point_maze_dataset.py

### DIFF
--- a/docs/tutorials/dataset_creation/point_maze_dataset.py
+++ b/docs/tutorials/dataset_creation/point_maze_dataset.py
@@ -308,11 +308,12 @@ for n_step in range(int(1e6)):
         print(n_step)
         if dataset is None:
             dataset = minari.create_dataset_from_collector_env(collector_env=collector_env,
-                                                               dataset_name=dataset_name,
-                                                               algorithm_name="QIteration",
-                                                               code_permalink="https://github.com/Farama-Foundation/Minari/blob/main/docs/tutorials/dataset_creation/point_maze_dataset.py",
-                                                               author="Rodrigo Perez-Vicente",
-                                                               author_email="rperezvicente@farama.org")
+            dataset_id=dataset_name,
+            algorithm_name="QIteration",
+            code_permalink="https://github.com/Farama-Foundation/Minari/blob/main/docs/tutorials/dataset_creation/point_maze_dataset.py",
+            author="Rodrigo Perez-Vicente",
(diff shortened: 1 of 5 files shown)


balisujohn (changes requested):
When I run `minari list remote`I get a KeyError  for `dataset_group.` Since this is not implemented yet for our existing datasets, it should just default to Unknown if there is no `dataset_group`key present in metadata.

 `file_size` is not included in metadata, so it needs to be obtained a different way. I'm not 100% sure how, but this might be helpful https://stackoverflow.com/questions/50875461/google-cloud-storage-get-object-size-api.`file_size` also needs to be manually obtained for local datasets from the local hdf5 files, since there isn't an existing field where it's explicitly kept track of. 
 
To get the environment name, you should be able to access  `json.loads(dst_metadata["env_spec"])["id"]`

minari/utils.py
```diff
@@ -323,6 +323,7 @@ def create_dataset_from_collector_env(
             "`author_email` is set to None. For longevity purposes it is highly recommended to provide an author email, or some other obvious contact information.",
             UserWarning,
         )
+
```
balisujohn, line 326: Looks like newlines added by accident.

docs/tutorials/dataset_creation/point_maze_dataset.py
```diff
@@ -308,11 +308,12 @@ def __call__(self, env, obs, info, action=None, rew=None, terminated=None, trunc
         print(n_step)
         if dataset is None:
             dataset = minari.create_dataset_from_collector_env(collector_env=collector_env,
-                                                               dataset_name=dataset_name,
-                                                               algorithm_name="QIteration",
-                                                               code_permalink="https://github.com/Farama-Foundation/Minari/blob/main/docs/tutorials/dataset_creation/point_maze_dataset.py",
-                                                               author="Rodrigo Perez-Vicente",
-                                                               author_email="rperezvicente@farama.org")
+            dataset_id=dataset_name,
```
balisujohn, line 311: good catch

younik:
Closing for inactivity; thanks a lot for doing this anyway!

PR #107, opened 2023-07-04
Minari version specifiers

rodrigodelazcano:
# Description

This PR adds the `minari_version` attribute to each dataset. This attribute is a version specifier which should contain the Minari versions that the dataset is compatible with (follows [PEP 440](https://peps.python.org/pep-0440/#version-specifiers)). When combining datasets `minari/utils.py::combine_minari_version_specifiers()` will calculate the intersection between all version specifiers for the resulting dataset. If there is no intersection, the datasets don't have any common minari dependency and can't be combined. The version specifier intersection adds a new dependency `portion` to create the version specifier intervals.

`minari_version` is specified when creating a dataset with `create_dataset_from_buffers` or `create_dataset_from_collector_env`. If not given a value, `minari_version` is initialized to `"=={version}"` where `version` is the local installed minari version.

This PR also contains some CLI updates:
- Bump `typer==0.9.0` and add `Annotated` to arguments and options as recommended by Typer docs.
- For the commands `minari list remote/local` only the compatible datasets with the local installed minari version and the latest version of each dataset will appear in the table. To show all dataset versions `--all` or `-a` can be used. Also the dataset names in the table are alphabetically ordered.
- Add `--force` option to `minari download dataset_id`.

Finally, used this PR to also remove Python 3.7 compatibility due to EOL. This has already been done for [Gymnasium](https://github.com/Farama-Foundation/Gymnasium/pull/573)

The PR is missing test which I'll add shortly. Current tests for downloading datasets are failing because the remote datasets need to be updated with the new attribute. I can proceed to update the remote datasets once the changes are approved.

# Checklist:

### DIFF
--- a/minari/__init__.py
+++ b/minari/__init__.py
@@ -38,4 +38,4 @@ __all__ = [
     "get_normalized_score",
 ]
 
-__version__ = "0.3.1"
+__version__ = "0.4.0"
--- a/minari/dataset/minari_storage.py
+++ b/minari/dataset/minari_storage.py
@@ -8,7 +8,8 @@ import gymnasium as gym
 import h5py
 import numpy as np
 from gymnasium.envs.registration import EnvSpec
-from packaging import version
+from packaging.specifiers import SpecifierSet
+from packaging.version import Version
 
 from minari.data_collector import DataCollectorV0
 from minari.serialization import deserialize_space
@@ -48,8 +49,8 @@ class MinariStorage:
             assert isinstance(minari_version, str)
 
             # Check that the dataset is compatible with the current version of Minari
-            assert version.parse(__version__) == version.parse(
-                minari_version
+            assert Version(__version__) in SpecifierSet(
+            minari_version
             ), f'Dataset {dataset_id} is compatible with Minari version {minari_version}. The Minari version of your system is {__version__}. Please install the appropriate version of Minari through : "pip install minari=={minari_version}'
             self._minari_version = minari_version
 
--- a/minari/dataset/minari_storage.py
+++ b/minari/dataset/minari_storage.py
@@ -11,6 +11,10 @@ from gymnasium.envs.registration import EnvSpec
 from minari.data_collector import DataCollectorV0
 from minari.serialization import deserialize_space
 
+import importlib.metadata
+
+# Use importlib due to circular import when: "from minari import __version__"
+__version__ = importlib.metadata.version('minari')
 
 PathLike = Union[str, bytes, os.PathLike]
 
@@ -38,6 +42,12 @@ class MinariStorage:
             dataset_id = f.attrs["dataset_id"]
             assert isinstance(dataset_id, str)
             self._dataset_id = dataset_id
+
+            minari_version = f.attrs["minari_version"]
+            assert isinstance(minari_version, str)
+
+            # Check that the dataset is compatible with the current version of Minari
+            assert version.parse(__version__) == version.parse(minari_version), f'Dataset {dataset_id} is compatible with Minari version {minari_version}. The Minari version of your system is {__version__}. Please install the appropiate version of Minari through : "pip install minari=={minari_version}'
 
             self._combined_datasets = f.attrs.get("combined_datasets", default=[])
 
--- a/minari/dataset/minari_storage.py
+++ b/minari/dataset/minari_storage.py
@@ -50,7 +50,7 @@ class MinariStorage:
 
             # Check that the dataset is compatible with the current version of Minari
             assert Version(__version__) in SpecifierSet(
-            minari_version
+                minari_version
             ), f'Dataset {dataset_id} is compatible with Minari version {minari_version}. The Minari version of your system is {__version__}. Please install the appropriate version of Minari through : "pip install minari=={minari_version}'
             self._minari_version = minari_version
 
--- a/minari/dataset/minari_storage.py
+++ b/minari/dataset/minari_storage.py
@@ -1,3 +1,4 @@
+import importlib.metadata
 import json
 import os
 from collections import OrderedDict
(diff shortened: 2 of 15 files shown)


minari/utils.py
```diff
@@ -1,20 +1,204 @@
...
+                if (
+                    combined_dataset_env_spec.max_episode_steps
+                    < dataset_env_spec.max_episode_steps
+                ):
+                    combined_dataset_env_spec.max_episode_steps = (
+                        dataset_env_spec.max_episode_steps
+                    )
+                else:
+                    dataset_env_spec.max_episode_steps = (
+                        combined_dataset_env_spec.max_episode_steps
```
younik, line 191: it can be done with `max`, isn't it?

minari/utils.py
```diff
@@ -1,20 +1,204 @@
...
+            raise ValueError(f"The dataset {dataset} is not of type MinariDataset.")
+        dataset_env_spec = dataset.spec.env_spec
+
+        assert isinstance(dataset_env_spec, EnvSpec)
+
+        # Look for the maximum value of `max_episode_steps` of the environments of all the
+        # datasets to combine and sets that value to the `max_episode_steps` of the combined dataset.
+        if combined_dataset_env_spec is None:
+            combined_dataset_env_spec = dataset_env_spec
+        elif dataset_env_spec.max_episode_steps is not None:
```
younik, line 176: If some of `dataset_env_spec.max_episode_steps` are None and some no, you ignore the ones with `None`. Is this the intended behavior? Or should we propagate the None?

And, why can a dataset max_episode_steps be None?

minari/utils.py
```diff
@@ -1,20 +1,204 @@
...
 from minari.dataset.minari_dataset import MinariDataset, clear_episode_buffer
 from minari.serialization import serialize_space
 from minari.storage.datasets_root_dir import get_dataset_path
 
 
+# Use importlib due to circular import when: "from minari import __version__"
+__version__ = importlib.metadata.version("minari")
+
+
+def combine_minari_version_specifiers(specifier_set: SpecifierSet):
```
younik, line 26: Do we need this function? Look like the & operator of SpecifierSet returns a valid SpecifierSet that we can use to check compatibility

younik (changes requested):
Overall looks good, I asked few comments

minari/utils.py
```diff
@@ -216,6 +380,17 @@ def create_dataset_from_buffers(
             "`author_email` is set to None. For longevity purposes it is highly recommended to provide an author email, or some other obvious contact information.",
             UserWarning,
         )
+    if minari_version is None:
+        warnings.warn(
+            f"`minari_version` is set to None. The compatible dataset version specifier for Minari will be automatically fixed to the installed version {__version__}.",
+            UserWarning,
+        )
+        minari_version = f"=={__version__}"
```
younik, line 388: maybe we want to support all the version with the first two numbers equal, i.e. for minari `0.3.1` we want `0.3.x` (?)

minari/dataset/minari_storage.py
```diff
@@ -190,3 +205,8 @@ def combined_datasets(self) -> List[str]:
     def id(self) -> str:
         """Name of the Minari dataset."""
         return self._dataset_id
+
+    @property
+    def minari_version(self) -> str:
+        """Version of Minari the dataset is compatible with."""
+        return self._minari_version
```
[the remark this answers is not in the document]
rodrigodelazcano: The functionality here will be to not provide a setter so that the user can't modify it.

minari/utils.py
rodrigodelazcano: Agree. `None` should be an indefinite number of steps so it should propagate. I'll fix this.

minari/utils.py
rodrigodelazcano: I agree. This makes more sense to be the default.

minari/dataset/minari_storage.py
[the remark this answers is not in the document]
younik: Oh you are right, I din't see it

minari/utils.py
rodrigodelazcano: done

rodrigodelazcano:
@younik I updated the code according to your comments. I'm missing a couple of tests for combining version specifiers

rodrigodelazcano:
Just added the tests. Updated `test_combine_datasets()` and created `test_combine_minari_version_specifiers()`

younik (approved):
LGTM, you can add the `minari_version` on remote datasets

balisujohn:
One other note, now that we are adding dataset versioning, we should remove the "Flatten Observation" and Flatten Actions" fields for the new version datasets, and additionally, the new version datasets should have the same data as the old datasets without flattening. For example: https://minari.farama.org/datasets/kitchen/complete/

minari/__init__.py
```diff
@@ -36,4 +36,4 @@
     "split_dataset",
 ]
 
-__version__ = "0.3.1"
+__version__ = "0.4.0"
```
rodrigodelazcano, line 39: We need to bump the minari version to the next release for the tests to download the compatible datasets with the current changes. This is the next Minari release correct? @balisujohn and @younik

rodrigodelazcano:
> One other note, now that we are adding dataset versioning, we should remove the "Flatten Observation" and Flatten Actions" fields for the new version datasets, and additionally, the new version datasets should have the same data as the old datasets without flattening. For example: https://minari.farama.org/datasets/kitchen/complete/

This PR should be ready to be merged. I just updated all the remote datasets:
1. The old dataset versions with the "flatten" attributes are pinned to the current minari release "==0.3.1".
2. The new unflatten datasets with the serialized spaces are compatible with minari ">0.3.1".

minari/__init__.py
balisujohn: I think it's fine for this to be the release PR. Should we merge https://github.com/Farama-Foundation/Minari/pull/111 first?

rodrigodelazcano: The test are missing but I think we can merge it now and I'll be working on the tests now.

younik: I think it is also fine to merge this PR (and bump the version even if it is not the last commit before releasing)

balisujohn:
LGTM :^)

PR #109, opened 2023-07-09
Update docs with EpisodeData info

enerrio:
# Description

This PR adds a section to the `Dataset Standards` documentation page describing the `EpisodeData` data structure and the fields it contains. Also includes a small snippet of sampling episodes from a Minari dataset.

## Type of change

# Checklist:

### DIFF
--- a/docs/content/dataset_standards.md
+++ b/docs/content/dataset_standards.md
@@ -540,3 +540,29 @@ The Minari storage format supports the following observation and action spaces:
 
 #### Space Serialization
 Spaces are serialized to a JSON format when saving to disk. This serialization supports all space types supported by Minari, and aims to be both human, and machine readable. The serialized action and observation spaces for the episodes in the dataset are saved as strings in the global HDF5 group metadata in `main_data.hdf5` for a particular dataset as `action_space` and `observation_space` respectively. All episodes in `main_data.hdf5` must have observations and actions that comply with these action and observation spaces. 
+
+## Minari Data Structures
+
+A Minari dataset is encapsulated in the `MinariDataset` class which allows for iterating and sampling through episodes which are defined as `EpisodeData` data class.
+
+### EpisodeData Structure
+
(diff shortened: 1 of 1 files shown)


docs/content/dataset_standards.md
```diff
@@ -540,3 +540,29 @@ The Minari storage format supports the following observation and action spaces:
...
+```
+
+The `sampled_episodes` variable will be a list of 10 `EpisodeData` elements, each containing episode data. An `EpisodeData` element is a data class consisting of the following fields:
+
+| Field             | Type         | Description                                                   |
+| ----------------- | ------------ | ------------------------------------------------------------- |
+| `id`              | `np.int64`   | ID of the episode.                                            |
+| `seed`            | `np.int64`   | Seed used to reset the episode.                               |
+| `total_timesteps` | `np.int64`   | Number of timesteps in the episode.                           |
+| `observations`    | `np.ndarray` | Observations for each timestep including initial observation. |
```
balisujohn, line 564: Observations and actions are not necessary represented as a single `np.ndarray`; it's worth looking at episodes from the various dummy envs in test_dataset_creation, to get an idea of what is possible for observations and action types.

balisujohn (changes requested):
One change, otherwise, looks good!

docs/content/dataset_standards.md
enerrio: @balisujohn I updated the data types with the test spaces I saw in the `test_dataset_creation.py` and `common.py` files. Let me know what you think.

balisujohn: I found the types `tuple`, `list` and `dict`, and `numpy.ndarray` from when I added at line 74 of test_dataset_creation.py: 
````
    print(type(dataset[0].observations))
````

It might be fair to say `List[str]` though if so, I'd confirm that string spaces are the only context where lists occur.

enerrio: I see, I updated data types. When it comes to `list` I only saw `List[str]` in datasets created in test_dataset_creation.py but left it as `list` in case there's some other dataset type I'm missing.

balisujohn (approved):

balisujohn:
LGTM :^)

PR #110, opened 2023-07-11
Get normalized scores

rodrigodelazcano:
# Description

This PR adds the [`get_normalized_score()` function from D4RL](https://github.com/Farama-Foundation/D4RL/blob/71a9549f2091accff93eeff68f1f3ab2c0e0a288/d4rl/offline_env.py#L71) for policy testing and comparisson. This function is added as a utility instead of a method in every MinariDataset since we haven't decided if we want to keep this feature for all future datasets in Minari and not all offline RL tasks can by tested in this manner (due to complexity of generating an expert policy for a reference maximum score).

This function requires the dataset to have two attributes which can be set when creating the dataset, `ref_min_score` and `ref_max_score`. These attributes are optional and if not present in the dataset `get_normalized_socre()` will through an error.
To set these attributes they can be directly passed by the user when creating the dataset, or by passing a callable expert policy to obtain the episode returns from. Note that if `ref_min_score` is not passed, it will be internally computed with a random policy.

I'm using this PR to also fix an issue with storing dictionary observation spaces in the DataCollectorV0 and not resetting the environment when `terminated` or `truncated`. I added a comment in the specific line changes.

This PR is a Draft because I need to update the PointMaze dataset generation tutorial to include the reference scores. There are some issues and differences with the D4RL PointMaze reference scores that I'll list here.

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/minari/data_collector/data_collector.py
+++ b/minari/data_collector/data_collector.py
@@ -447,7 +447,9 @@ class DataCollectorV0(gym.Wrapper):
         # Clear in-memory buffers
         self._buffer.clear()
 
-    def save_to_disk(self, path: str, dataset_metadata: Optional[Dict] = None):
+    def save_to_disk(
+        self, path: str, dataset_metadata: Optional[Dict[str, Any]] = None
+    ):
         """Save all in-memory buffer data and move temporary HDF5 file to a permanent location in disk.
 
         Args:
--- a/minari/data_collector/data_collector.py
+++ b/minari/data_collector/data_collector.py
@@ -176,7 +176,7 @@ class DataCollectorV0(gym.Wrapper):
                     assert isinstance(
                         episode_buffer[key], dict
                     ), f"Element to be inserted is type 'dict', but buffer accepts type {type(episode_buffer[key])}"
-                    
+
                     episode_buffer[key] = self._add_to_episode_buffer(
                         episode_buffer[key], value
                     )
@@ -231,7 +231,9 @@ class DataCollectorV0(gym.Wrapper):
         # certain conditions.
         if self._new_episode and not self._reset_called:
             if isinstance(self._previous_eps_final_obs, dict):
-                self._buffer[-1]["observations"] = self._add_to_episode_buffer({}, self._previous_eps_final_obs)
+                self._buffer[-1]["observations"] = self._add_to_episode_buffer(
+                    {}, self._previous_eps_final_obs
+                )
             else:
                 self._buffer[-1]["observations"] = [self._previous_eps_final_obs]
             if self._record_infos:
--- a/minari/data_collector/data_collector.py
+++ b/minari/data_collector/data_collector.py
@@ -176,6 +176,7 @@ class DataCollectorV0(gym.Wrapper):
                     assert isinstance(
                         episode_buffer[key], dict
                     ), f"Element to be inserted is type 'dict', but buffer accepts type {type(episode_buffer[key])}"
+                    
                     episode_buffer[key] = self._add_to_episode_buffer(
                         episode_buffer[key], value
                     )
@@ -203,7 +204,6 @@ class DataCollectorV0(gym.Wrapper):
             terminated=terminated,
             truncated=truncated,
         )
-
         # Force step data dictionary to include keys corresponding to Gymnasium step returns:
         # actions, observations, rewards, terminations, truncations, and infos
         assert STEP_DATA_KEYS.issubset(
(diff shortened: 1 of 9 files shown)


minari/data_collector/data_collector.py
```diff
@@ -230,7 +230,12 @@ def step(
         # or truncation. This may happen if the step_data_callback truncates or terminates the episode under
         # certain conditions.
         if self._new_episode and not self._reset_called:
-            self._buffer[-1]["observations"] = [self._previous_eps_final_obs]
+            if isinstance(self._previous_eps_final_obs, dict):
```
rodrigodelazcano, line 233: Fix for dictionary observation spaces

minari/utils.py
```diff
@@ -333,18 +414,76 @@ def create_dataset_from_collector_env(
...
+
+    Returns:
+        normalized_scores
+    """
+    with h5py.File(dataset.spec.data_path, "r") as f:
+        ref_min_score = f.attrs.get("ref_min_score", default=None)
+        ref_max_score = f.attrs.get("ref_max_score", default=None)
+    if ref_min_score is None or ref_max_score is None:
+        raise ValueError(
+            f"Reference score not provided for dataset {dataset.spec.dataset_id}. Can't compute the normalized score."
```
younik, line 483: Maybe it is better to access those attributes through `MinariStorage`
Should we make the `MinariDataset._data` gettable?

younik (approved):

minari/utils.py
```diff
@@ -197,6 +239,11 @@ def create_dataset_from_buffers(
         author (Optional[str], optional): author that generated the dataset. Defaults to None.
         author_email (Optional[str], optional): email of the author that generated the dataset. Defaults to None.
         code_permalink (Optional[str], optional): link to relevant code used to generate the dataset. Defaults to None.
+        ref_min_score( Optional[float], optional): minimum reference score from the average returns of a random policy. This value is later used to normalize a score with :meth:`minari.get_normalized_score`. If default None the value will be estimated with a default random policy.
```
Howuhh, line 242: should we even compute min score if max score is not provided? we won't be able to use it anyway

minari/utils.py
```diff
@@ -333,18 +414,76 @@ def create_dataset_from_collector_env(
...
+            env = copy.deepcopy(collector_env.env)
+            if ref_min_score is None:
+                ref_min_score = get_average_reference_score(
+                    env, RandomPolicy(env), num_episodes_average_score
+                )
+
+            if expert_policy is not None:
+                ref_max_score = get_average_reference_score(
+                    env, expert_policy, num_episodes_average_score
+                )
```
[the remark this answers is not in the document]
rodrigodelazcano: Sounds good, I'll update this section and then we can look into making other refactors

minari/utils.py
rodrigodelazcano: The idea here is to avoid extra optional attributes in the MinariStorage and MinariDataset. What we can do is move this function to the MinariStorage (that uses h5py`), and then call this function throught another method in MinariDataset. 
```python
def get_normalized_score(self, score):
    return self._data.get_normalized_score(score)
```

minari/utils.py
rodrigodelazcano: This is already being done. First checking if an `expert_policy` or `ref_max_score` is passed and then computing the respective attributes. I'll update the docs, I agree it's not very clear that functionality

minari/data_collector/data_collector.py
balisujohn: If possible, I think it would be good to add a test for this behavior.

rodrigodelazcano:
> Are we committing to at least leaving a version of `get_normalized_scores` in the codebase? If so, I think it would be good to write some sort of test for it. If not, then I think it's fine to hold off on tests for now.

We should be keeping it in all future codebase, at least for some of the d4rl datasets. I'll add the tests :)

minari/utils.py
younik: > The idea here is to avoid extra optional attributes in the MinariStorage and MinariDataset. What we can do is move this function to the MinariStorage (that uses h5py`), and then call this function throught another method in MinariDataset.
> 
> ```python
> def get_normalized_score(self, score):
>     return self._data.get_normalized_score(score)
> ```
> 
> Do you think this is a better option? In my opinion I prefer the current implementation to keep this feature as separated as possible from the dataset structures until we decide if we are going to maintain it at all.

I prefer the current implementation. I was also thinking on a general method `get_attr` to get some attribute from the data (and throw an error if there is no such attribute)

rodrigodelazcano:
I added a test for

> Are we committing to at least leaving a version of `get_normalized_scores` in the codebase? If so, I think it would be good to write some sort of test for it. If not, then I think it's fine to hold off on tests for now.

I added a test for this fix https://github.com/Farama-Foundation/Minari/pull/110#discussion_r1262846854. We can make the test for `get_normlized_score` in another PR

minari/utils.py
rodrigodelazcano: I like the idea of `get_attr` can you add this in a separe PR @younik?

balisujohn:
LGTM :^)

PR #102, opened 2023-06-28
ENH: improve MinariDataset

younik:
# Description

Add `__getitem__`and `__len__()` to `MinariDataset` and few refactors.

## Type of change

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [x] I have run `pytest -v` and no errors are present.
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #111, opened 2023-07-14
[Bug Fix] Fix h5py group creation

rodrigodelazcano:
# Description
While creating a dataset I got an error due to a simplification in this previous PR https://github.com/Farama-Foundation/Minari/pull/102 to create non-existing h5py groups. It appears that this code doesn't work as I thought. The following code is an example that we can't check for existing/non-existing groups and create the group as default with the `get` function in h5py:
```python
import h5py


with h5py.File('testing_file.hdf5', 'a') as f:

    f.create_group('new_group')

    group = f.get('new_group', default=f.create_group('new_group'))
```
The code will throw an error saying that the group already exists.

My proposed fix is a function to implement the previous functionality from the mentioned PR (https://github.com/Farama-Foundation/Minari/pull/102)

# Checklist:

### DIFF
--- a/minari/data_collector/data_collector.py
+++ b/minari/data_collector/data_collector.py
@@ -327,15 +327,6 @@ class DataCollectorV0(gym.Wrapper):
         self._reset_called = True
 
         return obs, info
-    
-    def _get_h5py_subgroup(self, group: h5py.Group, name: str) -> h5py.Group:
-        if name in group:
-            subgroup = group.get(name)
-            assert isinstance(subgroup, h5py.Group)
-        else:
-            subgroup = group.create_group(name)
-        
-        return subgroup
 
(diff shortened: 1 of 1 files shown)


balisujohn (commented):
Is there a test that could be added that would test this behavior? If so, I think that test should be added. Otherwise, LGTM

younik (approved):
Ah, my bad! The function is called even if the key exists; thanks for fixing this

rodrigodelazcano:
> Is there a test that could be added that would test this behavior? If so, I think that test should be added. Otherwise, LGTM

Lets add the test in another PR so that we can move forward with the next release of Minari

balisujohn:
LGTM :^)

PR #128, opened 2023-07-30
fix for action with noise type and value needs to be within (low,high)

### LINKED ISSUE #127: action can fall outside of action_space
https://github.com/Farama-Foundation/Minari/blob/da8578c8df55beb5f00f477f1f2d15d9741fbbe7/docs/tutorials/dataset_creation/point_maze_dataset.py#L358

```python 
>>> python pointmaze_dataset.py
/home/graham/micromamba/envs/ctx/lib/python3.10/site-packages/gymnasium/core.py:297: UserWarning: WARN: env.maze to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do `env.unwrapped.maze` for environment variables or `env.get_attr('maze')` that will search the reminding wrappers.
  logger.warn(
Traceback (most recent call last):
  File "/home/graham/code/inctxdt/compare/pointmaze/pointmaze_dataset.py", line 369, in <module>
    obs, rew, terminated, truncated, info = collector_env.step(action)
  File "/home/graham/micromamba/envs/ctx/lib/python3.10/site-packages/minari/data_collector/data_collector.py", line 216, in step
    assert self.dataset_action_space.contains(
AssertionError: Actions are not in action space.
```

also seems like there is an issue with the spaces .contain method whereby the assert in data_collector.py can fail if the dtype is float64 and the space dtype is float32

grahamannett:
# Description

Fixes #127 

either `action = np.clip(action, env.action_space.low, env.action_space.high, dtype=np.float32)` or `action = np.clip(action, -1, 1, dtype=np.float32)` should fix it

### DIFF
--- a/docs/tutorials/dataset_creation/point_maze_dataset.py
+++ b/docs/tutorials/dataset_creation/point_maze_dataset.py
@@ -356,6 +356,7 @@ for n_step in range(int(1e6)):
     action = waypoint_controller.compute_action(obs)
     # Add some noise to each step action
     action += np.random.randn(*action.shape)*0.5
+    action = np.clip(action, env.action_space.low, env.action_space.high, dtype=np.float32)
 
     obs, rew, terminated, truncated, info = collector_env.step(action)
     if (n_step + 1) % 200000 == 0:


balisujohn (approved):
LGTM :^)

PR #129, opened 2023-08-01
Improves MinariDataset load speed to speed up sampling, fixes total_steps bug and adds test coverage

balisujohn:
# Description

Previously, we would always iterate over all episodes to calculate total_steps when loading a MinariDataset. Now we will just load this from the MinariStorage when the MinariDataset is created, assuming no filter is specified. Empirically, this speeds up our sampling a lot because we no longer have to iterate over every episode regardless of how many we are sampling. 

Since the patch required changing how total_steps is loaded, I also extended test coverage for `total_steps` and found some bugs in how it's calculate in some settings, which this PR also patches. 

Performance of Minari v.s. Pickle for sampling 10 episodes across different dataset sizes.
image
Performance prior to patch:
image

## Type of change

# Checklist:


minari/dataset/minari_storage.py
```diff
@@ -247,10 +247,11 @@ def update_from_buffer(self, buffer: List[dict], data_path: str):
...
 
-            file.attrs.modify("total_episodes", last_episode_id + len(buffer))
-            file.attrs.modify(
-                "total_steps", file.attrs["total_steps"] + additional_steps
-            )
+            self._total_steps = file.attrs["total_steps"] + additional_steps
+            self._total_episodes = last_episode_id + len(buffer)
+
+            file.attrs.modify("total_episodes", self._total_episodes)
+            file.attrs.modify("total_steps", self._total_steps)
```
younik, line 254: Thanks for these fixes also. I am wondering if we need `self._total_steps` or if we can just make the `@property` total_steps reading from the file (and same for total_episodes)

minari/dataset/minari_dataset.py
```diff
@@ -141,18 +141,19 @@ def __init__(
...
         if episode_indices is None:
             episode_indices = np.arange(self._data.total_episodes)
+            total_steps = self._data.total_steps
+        else:
+            total_steps = sum(
+                self._data.apply(
+                    lambda episode: episode["total_timesteps"],
+                    episode_indices=episode_indices,
+                )
+            )
```
younik, line 151: This also means that creating subsets of the dataset can be very slow
Previously, I did a lazy initialization for total_steps attribute; we removed it to include it in the `MinariDatasetSpec`

younik (commented):
Not changes requests, just comments

minari/dataset/minari_dataset.py
balisujohn: We can maybe consider having metadata you can load without having to load the corresponding episode in a future PR, though I'm not super opinionated either way at this point.

minari/dataset/minari_storage.py
balisujohn: I think it might marginally save on file reads if we keep it the way it is for some use-cases, but it doesn't seem like it would make a super big difference either way.

younik (approved):

Issue #98, opened 2023-06-23
[Question] is hdf5 the best format for data storage?


Howuhh:
### Question

While hdf5 and h5py is the most popular approach for dimensional arrays storage, is has some major limitations. For example, the inability to read data in multiple processes / threads simultaneously, which can be important for the implementation of efficient data loading.

There is an alternative - [Zarr](https://zarr.readthedocs.io/en/stable/index.html), which is very similar, but a bit more capable. I think a discussion on this would be useful to the community.

elliottower:
Haven't tested it myself but it looks like hdf5 and h5py should be able to support multi-process reads (https://docs.h5py.org/en/latest/swmr.html?#multiprocess-concurrent-write-and-read) although multithreading seems to not be possible as far as I can see.

Haven't heard of zarr before but did some googling vs hdf5 and saw a few issues saying it was slower than hdf5 and this paper which seems to have that conclusion as well (but they don't mention multithreading besides once in the beginning so I'm guessing it's not testing that extensively, and that does seem to be one of the main advantages of zarr) https://arxiv.org/pdf/2207.09503.pdf 
As you say though Zarr does have advantages in concurrency and chunking which sounds useful (https://pythonspeed.com/articles/mmap-vs-zarr-hdf5/) Here's something comparing it with parquet https://sites.google.com/view/raybellwaves/blog/zarr-vs-parquet

We talked in the lambda meeting today a bit, the plan is to reach out to different people and see what they would prefer, to avoid arbitrarily changing it and changing to a new thing in the future. 

It seems like apache arrow may be a good choice: it's used by [huggingface datasets](https://huggingface.co/docs/datasets/v1.1.1/processing.html), as well as [ray data](https://docs.ray.io/en/latest/data/faq.html?highlight=arrow#what-is-arrow-and-how-does-ray-data-use-it) and as of  recently [pandas](https://datapythonista.me/blog/pandas-20-and-the-arrow-revolution-part-i). You can save them to disk as either Arrow/Feather files (uncompressed afaik but fast to read) or parquet files (compressed and more intended for long term storage). It looks like huggingface saves them directly as arrow files, so that seems like a reasonable thing to do here too imo. Converting from parquet and arrow is supposed to be very easy, and I think parquet would be flexible enough to support complex nested data like minari has but maybe not (more info https://arrow.apache.org/blog/2022/10/17/arrow-parquet-encoding-part-3/)

[...]

PR #133, opened 2023-08-23
Refactor DataCollectorV0 and HDF5 dependencies isolation

younik:
## Motivation

In the future, we aim to consider different storage methods than HDF5 files for our datasets (see https://github.com/Farama-Foundation/Minari/issues/98), as other backends can be faster.
Currently, different files depend on h5py; this PR isolates the dependency on MinariStorage. 
If we add support for other storage, we simply need to create a new MinariStorage that implements the same interface.

## Changes

To achieve the h5py isolation, we make DataCollectorV0, functions in utils and MinariDataset to interface with MinariStorage instead of directly to h5py. Thus, MinariStorage, should offer an API for the most used methods. 
This makes as natural dependencies `DataCollectorV0 -> MinariStorage.` and `MinariDataset -> MinariStorage`.

To avoid `MinariDataset -> DataCollectorV0` dependency, we change the current API for adding the buffer of a DataCollector to a MInariDataset, from:
```python
minari_dataset.update_from_env_collector(data_collector)
```
to:
```python
data_collector.add_to_dataset(minari_dataset)
```

Other user notable changes:
 - MinariStorage doesn't want anymore the path to the file, but the path to the directory containing the file(s).
 - The MinariStorage of a MinariDataset is now accessible using `.storage`
 - The init method of MinariStorage is for an existing dataset. To create a new storage, use the class method `new`
 - It is not possible anymore to combine dataset without copying. We may want to add this again in the future (but it will likely be incompatible with different backends).
  - Remove the custom error when env module is missing. There is already an informative error from Gymnasium (and if it is not informative enough, it should be changed there).


tests/utils/test_dataset_combine.py
```diff
@@ -173,32 +160,17 @@ def test_combine_datasets():
...
-    # testing with copy
-    combined_dataset = combine_datasets(
-        test_datasets, new_dataset_id="cartpole-combined-test-v0", copy=True
-    )
-    assert combined_dataset.spec.env_spec.max_episode_steps is None
-    _check_load_and_delete_dataset("cartpole-combined-test-v0")
-
     # Check that we get max(max_episode_steps) when there is no max_episode_steps=None
     test_datasets.pop()
-    # testing without creating a copy
```
balisujohn, line 192: What is the reason for removing these checks?

tests/dataset/test_minari_dataset.py
```diff
@@ -58,9 +62,9 @@ def test_episode_data(space: gym.Space):
 @pytest.mark.parametrize(
     "dataset_id,env_id",
     [
-        ("cartpole-test-v0", "CartPole-v1"),
-        ("dummy-dict-test-v0", "DummyDictEnv-v0"),
-        ("dummy-box-test-v0", "DummyBoxEnv-v0"),
+        # ("cartpole-test-v0", "CartPole-v1"),
```
balisujohn, line 65: I think these need to be uncommented

tests/utils/test_dataset_combine.py
younik: It is not possible anymore to have copy=False, as it was causing problem (the ids of episode were not consistent)

balisujohn (commented):
I noticed there is still an `h5py` import in utils.py and in it is used in `get_normalized_scores` Is this by design, and if so, what is blocking removing the h5py dependency there?

tests/utils/test_dataset_creation.py
```diff
@@ -50,15 +50,11 @@ def test_generate_dataset_with_collector_env(dataset_id, env_id):
...
     for episode in range(num_episodes):
-        terminated = False
-        truncated = False
-        while not terminated and not truncated:
+        done = False
+        while not done:
             action = env.action_space.sample()  # User-defined policy function
             _, _, terminated, truncated, _ = env.step(action)
-            if terminated or truncated:
-                assert not env._buffer[-1]
```
balisujohn, line 59: why is it safe to remove these asserts?

tests/utils/test_dataset_combine.py
younik: It is not specific about h5py isolation, but it is a bug in the current code: when you combine two datasets, the episode ids of the second dataset are modified when you have copy=False, i.e. they don't start anymore from 0

tests/utils/test_dataset_creation.py
younik: I changed the internal structure and the assert is not true anymore (when resetted, the buffer is not void, but immediately has the episode id)
Anyway, it wasn't a meaningful test (it doesn't test any wanted feature)

younik:
> 

> I noticed there is still an `h5py` import in utils.py and in it is used in `get_normalized_scores` Is this by design?

Oh, thanks for spotting this, it wasn't by design; I fixed it.

tests/common.py
```diff
@@ -547,7 +549,7 @@ def check_load_and_delete_dataset(dataset_id: str):
 
 
 def create_dummy_dataset_with_collecter_env_helper(
-    dataset_id: str, env: DataCollectorV0, num_episodes: int = 10
+    dataset_id: str, env: DataCollectorV0, num_episodes: int = 10, **kwargs
```
balisujohn, line 552: what is the purpose of the `kwargs`?

balisujohn (commented):
Just had one more question, otherwise, once this is passing pre-commit, I think it's ready to merge :^)

tests/common.py
younik: It is needed for this new test; they are a general kwarg that can be passed to the`create_dataset_from_collector_env` function  https://github.com/Farama-Foundation/Minari/blob/c5c88e8a03921ac8012a632c741916560670acc2/tests/utils/test_get_normalized_score.py#L20-L27

balisujohn (approved):
Ready to merge :^) but one last question, what it two datasets that are themselves each the result of combining two datasets are combined? Will this situation be handled correctly?

younik:
> Ready to merge :^) but one last question, what it two datasets that are themselves each the result of combining two datasets are combined? Will this situation be handled correctly?

Thanks! Yes, this should not create any issue as the new generated dataset is a perfectly normal dataset (everything is copied); with just different metadata.

PR #137, opened 2023-08-31
make env optional arg while creating from buffers

avjmachine:
# Description

Make the env argument as optional in the create_dataset_from_buffers() function in utils

## Type of change

### Screenshots

# Checklist:


balisujohn (changes requested):
Thanks for your work on this so far. I requested changes in two places. I also think there should be a new test which creates a dataset without providing an `env` argument, and makes sure that everything still works as intended. Once that is all done, I will review again.

balisujohn:
Thanks for your work on this; it looks almost ready! I think it would be good to add a test to test_dataset_creation where a dataset is created from buffers without the env being set, and then also tries to load that dataset after it is created. Additionally, it looks like pre-commit is failing. Once those two things are done, it should be ready to merge.

minari/utils.py
```diff
@@ -433,10 +433,16 @@ def create_dataset_from_buffers(
...
+    if env is None:
+        if observation_space is None or action_space is None:
+            raise ValueError(
+                             "Both observation space and action space must be provided, if environment is None"
+            )
+    else:
+        if observation_space is None:
+            observation_space = env.observation_space
+        if action_space is None:
+            action_space = env.action_space
```
younik, line 445: can you reduce the complexity of this making a single if for raising the error (then you don't need the else)

minari/utils.py
```diff
@@ -353,8 +353,8 @@ def get_average_reference_score(
 
 def create_dataset_from_buffers(
     dataset_id: str,
-    env: gym.Env,
     buffer: List[Dict[str, Union[list, Dict]]],
+    env: Optional[gym.Env] = None,
```
younik, line 357: notice that this is a breaking change (positional calling will not work anymore), we may want to delay this for minari 0.5, that may be the next release anyway

avjmachine: Okay, will work on the complexity (if conditions). 
Yes, it is a breaking change. I will keep the changes ready in any case.

I also had a question regarding the test case:

Which of the following options do we use for the env buffers in the absence of an env?
a. any buffers already available. I don't know if there is any, do you?
b. create buffer data by running one of the gym env
c. Manually create a dummy env buffer object

younik: What do you mean with env buffers? We have a field in the metadata to save the env_spec; in case env is missing we will miss that attribute in the metadata. The env can missing, for example if the data comes from a real robot with no associated simulated environment. Does this answer your question?

avjmachine: Let me rephrase my question. What do we pass as observation space and action space when env is None for the test case? Which one if those 3 options listed would be better?

younik: If env is not specified, observation_space and action_space must be specified
This is what the if branch of lines 436-440 does, right?

younik: The user must specify it

avjmachine: Yes, the user must specify it.
That's why I was wondering what to use for the "test case". I wanted to check if we already have some action space observation space data for the test case, or should I create one. I listed those 3 options to get your advice on what would be the best option.

younik: Oh clear now, you are asking it for tests
For that you can follow the same workflow of this: https://github.com/Farama-Foundation/Minari/blob/c43a61249ac51dd663a12983b7050aced97115b0/tests/utils/test_dataset_creation.py#L112

So creating the buffer from env (they are test env testing different spaces; I don't think it is necessary to test that function, one is enough). If you need to copy-paste the code from that test, feel free to create a method in tests/common.py with a function to create the buffer from an env

avjmachine: Yes, thanks a lot, I was asking about the test. In the absence of env, we need to pass action space or observation space mandatorily. Hence, I wanted ideas on what arguments to pass for observation space and action space parameters when we test this scenario.

younik: You should use the same observation_space and action_space of the environment that generates the dataset (`env.observation_space` and env.action_space`), then, for the sake of the test, you don't pass the environment but only these spaces

minari/utils.py
avjmachine: Made the changes.

minari/utils.py
```diff
@@ -482,18 +483,22 @@ def create_dataset_from_buffers(
...
+                    "env_spec"
+                ] = env.spec.to_json()  # pyright: ignore [reportOptionalMemberAccess]
             file.attrs["dataset_id"] = dataset_id
 
-            action_space_str = serialize_space(action_space)
-            observation_space_str = serialize_space(observation_space)
-
-            file.attrs["action_space"] = action_space_str
-            file.attrs["observation_space"] = observation_space_str
+            if action_space is not None:
```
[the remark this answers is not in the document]
avjmachine: Changes are done

minari/utils.py
```diff
@@ -433,10 +433,11 @@ def create_dataset_from_buffers(
     except InvalidSpecifier:
         print(f"{minari_version} is not a version specifier.")
 
-    if observation_space is None:
-        observation_space = env.observation_space
-    if action_space is None:
-        action_space = env.action_space
+    if env is not None:
```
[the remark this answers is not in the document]
avjmachine: Changes are done

minari/utils.py
avjmachine: Have resolved the complexity for if while raising error and suppressed pyright errors. Working on the test case.

avjmachine: Have resolved the complexity for if while raising error and suppressed pyright errors. Working on the test case.

minari/utils.py
```diff
@@ -172,7 +172,8 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
         max_episode_steps = None
     else:
         max_episode_steps = max(
-            dataset.spec.env_spec.max_episode_steps for dataset in datasets_to_combine
+            dataset.spec.env_spec.max_episode_steps
+            for dataset in datasets_to_combine  # pyright: ignore[reportGeneralTypeIssues]
```
younik, line 176: Actually, since now the datasets can have a env_spec=None, we should address also this case.
You can create the list of env_spec and filter out None values

minari/utils.py
```diff
@@ -292,7 +293,9 @@ def get_average_reference_score(
     for _ in range(num_episodes):
         while True:
             action = policy(obs)
-            obs, _, terminated, truncated, info = env.step(action)
+            obs, _, terminated, truncated, info = env.step(
+                action  # pyright: ignore[reportGeneralTypeIssues]
+            )
```
younik, line 298: What was the warning here? I cannot see it

minari/utils.py
```diff
@@ -394,10 +397,17 @@ def create_dataset_from_buffers(
...
 
+    if env is None and (observation_space is None or action_space is None):
+        raise ValueError(
+            "Both observation space and action space must be provided, if environment is None"
+        )
+
     if observation_space is None:
-        observation_space = env.observation_space
+        observation_space = (
+            env.observation_space  # pyright: ignore[reportOptionalMemberAccess]
```
younik, line 407: why do you need this? There is no way reaching here when env is None; do you got errors from the pre-commit?

minari/utils.py
```diff
@@ -394,10 +397,17 @@ def create_dataset_from_buffers(
...
+        )
     if action_space is None:
-        action_space = env.action_space
+        action_space = env.action_space  # pyright: ignore[reportOptionalMemberAccess]
```
younik, line 410: same as above

tests/utils/test_dataset_combine.py
```diff
@@ -72,8 +72,8 @@ def _generate_dataset_with_collector_env(
         # Force None max_episode_steps
         env_spec = gym.make("CartPole-v1").spec
         assert env_spec is not None
-        env_spec.max_episode_steps = None
-        env = env_spec.make()
+        env_spec.max_episode_steps = None  # pyright: ignore[reportOptionalMemberAccess]
+        env = env_spec.make()  # pyright: ignore[reportOptionalMemberAccess]
```
younik, line 76: env_space is not None for sure, I think you can remove the pyright ignore

minari/utils.py
avjmachine: As of now, I'm raising an error if any of the datasets to be combined have no env_spec. Please do let me know if you'd prefer some other specific action instead of this.

minari/utils.py
avjmachine: It was the pyright raising error for reporting General Type Issue while running pre-commit hook.

minari/utils.py
avjmachine: Yes, I got errors from pre-commit.

minari/utils.py
avjmachine: Yes , pyright error for optional member access.

tests/utils/test_dataset_combine.py
avjmachine: I've removed all pyright inline ignores. Please do let me know if I should set the pyright ignore in the .toml file or if there is some other option to resolve this.

younik (commented):
I had to play around a bit to understand how to solve so I just pushed the code to your branch; now should be good. 
Last thing before merging, can you add a test for combining datasets without env_spec?

avjmachine:
Okay, so using assert you could handle the pyright optional member access issue! Thanks a lot for helping with that!

minari/utils.py
```diff
@@ -161,31 +163,27 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
...
-        combine_env_spec.append(dataset_env_spec)
-
-    assert all(
-        env_spec == combine_env_spec[0] for env_spec in combine_env_spec
-    ), "The datasets to be combined have different values for `env_spec` attribute."
+        assert isinstance(dataset, MinariDataset)
+        env_spec = dataset.spec.env_spec
+        if env_spec is not None:
+            assert (
+                common_env_spec is not None
```
younik, line 174: Oh my bad, you are right, thanks! Can you change this assert to check env_spec == common_env_spec? This should solve the issue

Would be nice to also add a test on combining with different env_spec then (an error is expected)

avjmachine: okay, sure. I'll add the fix and the test as well.

minari/utils.py
```diff
@@ -161,31 +163,27 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
...
+            ), "Found incompatible env_spec in datasets"
+            if (
+                common_env_spec.max_episode_steps is None
+                or env_spec.max_episode_steps is None
+            ):
+                common_env_spec.max_episode_steps = None
+            else:
+                common_env_spec.max_episode_steps = max(
+                    common_env_spec.max_episode_steps, env_spec.max_episode_steps
+                )
```
younik, line 184: this sounds good, but it still doesn't check if the two env_spec are the same.
One thing you can do is creating a copy of env_spec, modify the max_episode_steps to be the same of common_env_spec and then check env_spec == common_env_spec

tests/common.py
```diff
@@ -403,6 +403,9 @@ def check_env_recovery_with_subset_spaces(
         observation_space_subset (gym.spaces.Space): desired subset observation space
 
     """
+    if dataset.env_spec is None:
+        raise ValueError("Recovering environment is not possible when env_spec is None")
```
younik, line 409: this it is not necessary I believe, as `recover_environment` will throw an error itself if env_spec is None

minari/utils.py
younik: Thanks for the fix! I left a couple of comments.
You are right about the assert, the difference is: when you believe something cannot happen you can use assert (e.g. asserts on line 401 and 404 of utils.py, as there is the check above). When it is something that can happen, but you want to let the user that it is not allowed, you should raise an error.
Unfortunately, in Minari we often use asserts as exceptions. I would be happy to merge another PR that addresses it.

> Edit: Just noticed that we need a separate for loop statement for updating the max_episode_steps after line 186. I missed that. Will correct it. But please do advise me on the assert statement.

What do you mean? I don't see it.

minari/utils.py
avjmachine: yes, that's correct. I'm doing the same in lines after 184. (the edit comment that I mentioned)

avjmachine:
@younik I'm currently working on the merge conflicts with the eval_env changes that got merged yesterday. I had one question regarding this. Could there be a scenario where there is an eval_env even when env is None? I mean, can a user provide a eval_env for a dataset generated with buffers without an env?

younik:
> @younik I'm currently working on the merge conflicts with the eval_env changes that got merged yesterday. I had one question regarding this. Could there be a scenario where there is an eval_env even when env is None? I mean, can a user provide a eval_env for a dataset generated with buffers without an env?

Thank you, good point. I don't see any common case when this would happen, but also no reason to constraint it. I would simply do what is easier in code, which is no constraint I guess.

minari/utils.py
```diff
@@ -494,11 +508,24 @@ def create_dataset_from_buffers(
...
     elif isinstance(env, gym.Env):
         env_spec = env.spec
+    elif env is None:
+        env_spec = None
     else:
-        raise ValueError("The `env` argument must be of types str|EnvSpec|gym.Env")
+        raise ValueError("The `env` argument must be of types str|EnvSpec|gym.Env|None")
+
+    if isinstance(env, (str, EnvSpec, type(None))) and (observation_space is None or action_space is None):
+        raise ValueError("Both observation space and action space must be provided, if env is str|EnvSpec|None")
```
avjmachine, line 517: I'm assuming here that action_space and observation_space cannot be extracted unless it is of type gym.Env

minari/utils.py
avjmachine: Added the fix.

minari/utils.py
avjmachine: So, if it is a str or EnvSpec, there would be 2 steps to extract the action_space and observation_space? 

1. recover env as gym.Env with gym.make; 
2. recover action_space and observation_space from this gym.Env.

Is this correct?

younik: Yes, exactly

tests/common.py
avjmachine: removed the redundant raise error

minari/utils.py
```diff
@@ -516,18 +516,18 @@ def create_dataset_from_buffers(
...
         env_spec = None
     else:
         raise ValueError("The `env` argument must be of types str|EnvSpec|gym.Env|None")
 
-    if isinstance(env, (str, EnvSpec, type(None))) and (observation_space is None or action_space is None):
-        raise ValueError("Both observation space and action space must be provided, if env is str|EnvSpec|None")
-
+    if isinstance(env, (str, EnvSpec)):
+        env = gym.make(env)
+    assert env is not None
```
avjmachine, line 527: need this redundant statement to avoid pyright error

minari/utils.py
avjmachine: implemented the above logic

minari/utils.py
```diff
@@ -464,10 +485,10 @@ def create_dataset_from_buffers(
 
     Args:
         dataset_id (str): name id to identify Minari dataset.
-        env (str|gym.Env|EnvSpec): Gymnasium environment(gym.Env)/environment id(str)/environment spec(EnvSpec) used to collect the buffer data.
         buffer (list[Dict[str, Union[list, Dict]]]): list of episode dictionaries with data.
+        env (Optional[str|gym.Env|EnvSpec]): Gymnasium environment(gym.Env)/environment id(str)/environment spec(EnvSpec) used to collect the buffer data. Defaults to None.
         eval_env (Optional[str|gym.Env|EnvSpec]): Gymnasium environment(gym.Env)/environment id(str)/environment spec(EnvSpec) to use for evaluation with the dataset. After loading the dataset, the environment can be recovered as follows: `MinariDataset.recover_environment(eval_env=True).
-                                                If None the `env` used to collect the buffer data should be used for evaluation.
+                                                If None, and if the `env` used to collect the buffer data is available, latter should be used for evaluation.
```
younik, line 491: not related to your PR, but `should` -> `will`

minari/utils.py
```diff
@@ -161,31 +163,44 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
...
 
-    combine_env_spec = []
+    # updating the common_env_spec's max_episode_steps
     for dataset in datasets_to_combine:
-        dataset_env_spec = copy.deepcopy(dataset.spec.env_spec)
-        dataset_env_spec.max_episode_steps = max_episode_steps
-        combine_env_spec.append(dataset_env_spec)
+        assert isinstance(dataset, MinariDataset)
+        env_spec = dataset.spec.env_spec
+        if env_spec is not None:
```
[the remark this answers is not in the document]
avjmachine: okay, so if we have just 2 types of datasets - one type as None and other with an env_spec, then we throw an error? Or do we filter out the None types and just combine the others with a common env_spec?

minari/utils.py
```diff
@@ -161,31 +163,44 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
...
+                common_env_spec.max_episode_steps = max(
+                    common_env_spec.max_episode_steps, env_spec.max_episode_steps
+                )
 
-    assert all(
-        env_spec == combine_env_spec[0] for env_spec in combine_env_spec
-    ), "The datasets to be combined have different values for `env_spec` attribute."
+    # checking equivalence of all datasets with an env
+    for dataset in datasets_to_combine:
+        if dataset.spec.env_spec is not None:
```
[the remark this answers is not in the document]
avjmachine: okay, I think this answers my question in the previous comment.

minari/utils.py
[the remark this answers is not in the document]
younik: We should throw an error

minari/utils.py
avjmachine: updated

minari/utils.py
```diff
@@ -392,11 +407,17 @@ def _generate_dataset_metadata(
...
                   If None the environment used to collect the data (`env={env_spec}`) will be used for this purpose.",
-            UserWarning,
-        )
+                UserWarning,
+            )
+        else:
+            warnings.warn(
+                "Since both `eval_env` and `env` used to collect the data are None, no environment can be used for evaluation",
+                UserWarning,
+            )
```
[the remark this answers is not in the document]
avjmachine: made the 2 checks independent

minari/utils.py
```diff
@@ -161,31 +163,44 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
...
+    # checking equivalence of all datasets with an env
+    for dataset in datasets_to_combine:
+        if dataset.spec.env_spec is not None:
+            # updating max_episode_steps to a common value for sake of checking equality
+            env_spec_copy = copy.deepcopy(dataset.spec.env_spec)
+            env_spec_copy.max_episode_steps = common_env_spec.max_episode_steps
+            if env_spec_copy != common_env_spec:
+                raise ValueError(
+                    "The datasets to be combined have different values for `env_spec` attribute."
+                )
```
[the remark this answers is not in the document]
avjmachine: done

minari/utils.py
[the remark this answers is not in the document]
avjmachine: updated

minari/utils.py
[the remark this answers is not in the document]
avjmachine: changes implemented

younik (approved):
LGTM; thanks for implementing this!

younik:
Actually, can you fix pre-commit @avjmachine 

Looks like you should use `env_spec` in line 527 & 530 in utils, as env can be just a string.
For the documentation, it is not related to your PR, so you can ignore it

avjmachine:
> Actually, can you fix pre-commit @avjmachine
> 
> Looks like you should use `env_spec` in line 527 & 530 in utils, as env can be just a string. For the documentation, it is not related to your PR, so you can ignore it

In line 524, we are overwriting the env variable if its just a string, using "env = gym.make(env)". So, will this work or is there a risk of env not having an observation_space or action_space attribute even after using gym.make?

younik:
Yes, it is

> > Actually, can you fix pre-commit @avjmachine
> > Looks like you should use `env_spec` in line 527 & 530 in utils, as env can be just a string. For the documentation, it is not related to your PR, so you can ignore it
> 
> In line 524, we are overwriting the env variable if its just a string, using "env = gym.make(env)". So, will this work or is there a risk of env not having an observation_space or action_space attribute even after using gym.make?

No, there isn't; however, pyright is not able to infer it and throws an error.
Changing `assert env is not None` to `assert isinstance(env, gym.Env)` should work

avjmachine:
> Yes, it is
> 
> > > Actually, can you fix pre-commit @avjmachine
> > > Looks like you should use `env_spec` in line 527 & 530 in utils, as env can be just a string. For the documentation, it is not related to your PR, so you can ignore it
> > 
> > 
> > In line 524, we are overwriting the env variable if its just a string, using "env = gym.make(env)". So, will this work or is there a risk of env not having an observation_space or action_space attribute even after using gym.make?
> 
> No, there isn't; however, pyright is not able to infer it and throws an error. Changing `assert env is not None` to `assert isinstance(env, gym.Env)` should work

Done. I didn't get this pyright error on my local env. Hence, got confused.
Also, fixed additional pyright errors which I got while committing. You can review those changes. I changed int64 to integer for type check based on info [here](https://github.com/numpy/numpy/issues/2951) and [here](https://github.com/microsoft/pylance-release/issues/3830)

PR #143, opened 2023-09-09
added dataset size attribute to minari dataset

shreyansjainn:
# Description

added dataset size attribute to minari dataset generation function and cli

## Type of change

### Screenshots

# Checklist:


balisujohn:
Thanks for your contribution!

So a few comments, `get_dataset_size` should be a public function of MinariStorage if possible, since it's not backend agnostic.

I created a local dataset and called `minari list local` it displayed a number, but was missing MB after to specify units, so that would be a good thing to add. 

Also, I think it would be good to add a test for `get_dataset_size`. Basically just a small new test in `test_minari_storage` which creates a small dataset and makes sure it is the same size on file as the size returned by get_dataset_size. 


Let me know if you have any comments or questions about the changes I requested, otherwise, this looks good and should correctly display the dataset sizes once we have the hosted datasets annotated with them.

PR #144, opened 2023-09-10
small simple bug-fix: update lost infos in function create_dataset function

im-Kitsch:
# Description

This is fix for simple bug-fix for function `create_dataset_from_collector_env()`.  
1. Add docstring for parameters
2. Add metadata attribute to dataset, meta data like `algorithm name` `author` `author Email` is not saved though it's passed into function.

for function `create_dataset_from_collector_env()`
ref_min/ref_max score is not saved to metadata, it's easily to find the problem.

I think the bugs are quite clear and definitely is a bug, you can take a look,

## Type of change

# Checklist:

### DIFF
--- a/minari/utils.py
+++ b/minari/utils.py
@@ -489,10 +489,10 @@ def create_dataset_from_buffers(
                 "env_spec"
             ] = env.spec.to_json()  # pyright: ignore [reportOptionalMemberAccess]
             file.attrs["dataset_id"] = dataset_id
-            file.attrs["algorithm_name"] = algorithm_name
-            file.attrs["author"] = author
-            file.attrs["author_email"] = author_email
-            file.attrs["code_permalink"] = code_permalink
+            file.attrs["algorithm_name"] = str(algorithm_name)
+            file.attrs["author"] = str(author)
+            file.attrs["author_email"] = str(author_email)
+            file.attrs["code_permalink"] = str(code_permalink)
 
             action_space_str = serialize_space(action_space)
             observation_space_str = serialize_space(observation_space)
@@ -613,6 +613,7 @@ def create_dataset_from_collector_env(
             "author": str(author),
             "author_email": str(author_email),
             "code_permalink": str(code_permalink),
+            "minari_version": minari_version,
         }
 
         if expert_policy is not None or ref_max_score is not None:
@@ -636,14 +637,7 @@ def create_dataset_from_collector_env(
 
         collector_env.save_to_disk(
             data_path,
-            dataset_metadata={
-                "dataset_id": str(dataset_id),
-                "algorithm_name": str(algorithm_name),
-                "author": str(author),
-                "author_email": str(author_email),
-                "code_permalink": str(code_permalink),
-                "minari_version": minari_version,
-            },
+            dataset_metadata=dataset_metadata,
         )
         return MinariDataset(data_path)
     else:
--- a/minari/utils.py
+++ b/minari/utils.py
@@ -289,10 +289,6 @@ def combine_datasets(
                         combined_data_file.attrs.modify(
                             optional_parameter, dataset_file.attrs[optional_parameter]
                         )
-                # combined_data_file.attrs.modify("author", dataset_file.attrs["author"])
-                # combined_data_file.attrs.modify(
-                #     "author_email", dataset_file.attrs["author_email"]
-                # )
 
         assert combined_dataset_env_spec is not None
         combined_data_file.attrs["env_spec"] = combined_dataset_env_spec.to_json()
@@ -637,10 +633,6 @@ def create_dataset_from_collector_env(
         data_path = os.path.join(dataset_path, "main_data.hdf5")
         dataset_metadata: Dict[str, Any] = {
             "dataset_id": str(dataset_id),
-            # "algorithm_name": str(algorithm_name),
-            # "author": str(author),
-            # "author_email": str(author_email),
-            # "code_permalink": str(code_permalink),
             "minari_version": minari_version,
         }
 
--- a/minari/utils.py
+++ b/minari/utils.py
@@ -503,10 +503,6 @@ def create_dataset_from_buffers(
                 "env_spec"
             ] = env.spec.to_json()  # pyright: ignore [reportOptionalMemberAccess]
             file.attrs["dataset_id"] = dataset_id
-            # file.attrs["algorithm_name"] = str(algorithm_name)
-            # file.attrs["author"] = str(author)
-            # file.attrs["author_email"] = str(author_email)
-            # file.attrs["code_permalink"] = str(code_permalink)
 
             if algorithm_name is not None:
                 file.attrs["algorithm_name"] = str(algorithm_name)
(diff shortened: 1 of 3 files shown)


balisujohn:
Good catches. Thanks for your contribution! All these fixes look correct to me.  I think it would be good to make the values for `author_name`, `author_email`, `code_permalink`  default to `"not provided"` instead of `str(None)` if not provided. Once that's done I will merge this PR after a final review.  As for minari version as an argument, I'm wondering if we should make it non-optional, but we can sort that out in a future PR.

im-Kitsch:
Hi, 

I pushed update to save `author_name`, `author_email`, `code_permalink` and `algorithm name` only when it's not None.  And additionally I changed a code of combine dataset so that it could optional parameters are not saved. 

I agree that we need consider if minari version should be optional or mandatory. 

Futhermore, I think the dataset's structure should be **clearly** specified and let it be **stable**. 

E.g. here in `combine_dataset`, if `author` or `author_email` are not saved then it will throw error. 
https://github.com/Farama-Foundation/Minari/blob/ea4097854cae1b333ddd7de2fc3fabd0fbc1076d/minari/utils.py#L277-L290

There should be still small errors like this one.

minari/utils.py
```diff
@@ -284,10 +284,15 @@ def combine_datasets(
...
-                combined_data_file.attrs.modify("author", dataset_file.attrs["author"])
-                combined_data_file.attrs.modify(
-                    "author_email", dataset_file.attrs["author_email"]
-                )
+                for optional_parameter in ['author', 'author_email', 'code_permalink', 'algorithm_name']:
+                    if optional_parameter in dataset_file.attrs:
+                        combined_data_file.attrs.modify(
+                            optional_parameter, dataset_file.attrs[optional_parameter]
+                        )
+                # combined_data_file.attrs.modify("author", dataset_file.attrs["author"])
```
balisujohn, line 292: please remove commented out lines

balisujohn (changes requested):
The code seems almost ready to merge. Please remove the commented out lines, and add a basic test to for behavior when the  optional `dataset_metadata` keys modified in this PR are provided when creating an instance of `MinariDataset`

im-Kitsch:
Hi, 
I added one test case for `combine dataset` and modified `create dataset` test case, could you take a look? 

I think these tests should be modified in the future again, since currently those additional infomations could not be got directly from `Minaridataset` instance. But have to be read manualy from h5py file.

minari/utils.py
```diff
@@ -486,6 +503,19 @@ def create_dataset_from_buffers(
                 "env_spec"
             ] = env.spec.to_json()  # pyright: ignore [reportOptionalMemberAccess]
             file.attrs["dataset_id"] = dataset_id
+            # file.attrs["algorithm_name"] = str(algorithm_name)
+            # file.attrs["author"] = str(author)
+            # file.attrs["author_email"] = str(author_email)
+            # file.attrs["code_permalink"] = str(code_permalink)
```
younik, line 509: can you remove these comments?

tests/utils/test_dataset_combine.py
```diff
@@ -231,3 +232,53 @@ def test_combine_minari_version_specifiers(specifier_intersection, version_speci
...
+                action = env.action_space.sample()  # User-defined policy function
+                _, _, terminated, truncated, _ = env.step(action)
+            env.reset()
+
+        # Create Minari dataset and store locally
+        dataset = minari.create_dataset_from_collector_env(
+            dataset_id=dataset_id,
+            collector_env=env,
+            algorithm_name="random_policy" + str(i),
+            code_permalink="https://github.com/Farama-Foundation/Minari/blob/f095bfe07f8dc6642082599e07779ec1dd9b2667/tutorials/LocalStorage/local_storage.py"
```
younik, line 259: can you change the permalink to be the following? https://github.com/Farama-Foundation/Minari/blob/main/tests/utils/test_dataset_combine.py

tests/utils/test_dataset_combine.py
```diff
@@ -231,3 +232,53 @@ def test_combine_minari_version_specifiers(specifier_intersection, version_speci
...
+        dataset_list.append(dataset)
+
+    combined_dataset = combine_datasets(
+        dataset_list, new_dataset_id="cartpole-combined-test-v0"
+    )
+    with h5py.File(combined_dataset.spec.data_path) as dt_file:
+        assert dt_file.attrs["algorithm_name"] == "random_policy" + str(n_data - 1)
+        _final_code_link = (
+            "https://github.com/Farama-Foundation/Minari/blob/f095bfe07f8dc6642082599e07779ec1dd9b2667/tutorials/LocalStorage/local_storage.py"
+            + str(n_data - 1)
```
younik, line 275: same here

tests/utils/test_dataset_creation.py
```diff
@@ -72,6 +73,14 @@ def test_generate_dataset_with_collector_env(dataset_id, env_id):
         author_email="wdudley@farama.org",
     )
 
+    # test metadata
+    with h5py.File(dataset.spec.data_path, "r") as data_file:
+        assert data_file.attrs["algorithm_name"] == "random_policy"
+        code_link = "https://github.com/Farama-Foundation/Minari/blob/f095bfe07f8dc6642082599e07779ec1dd9b2667/tutorials/LocalStorage/local_storage.py"
```
younik, line 79: same here

tests/utils/test_dataset_creation.py
```diff
@@ -279,6 +288,14 @@ def test_generate_dataset_with_space_subset_external_buffer():
         observation_space=observation_space_subset,
     )
 
+    # test metadata
+    with h5py.File(dataset.spec.data_path, "r") as data_file:
+        assert data_file.attrs["algorithm_name"] == "random_policy"
+        code_link = "https://github.com/Farama-Foundation/Minari/blob/f095bfe07f8dc6642082599e07779ec1dd9b2667/tutorials/LocalStorage/local_storage.py"
```
younik, line 294: same ehre

tests/utils/test_dataset_combine.py
```diff
@@ -231,3 +232,53 @@ def test_combine_minari_version_specifiers(specifier_intersection, version_speci
...
+        )
+        assert dt_file.attrs["code_permalink"] == _final_code_link
+        assert dt_file.attrs["author"] == "WillDudley" + str(n_data - 1)
+        assert dt_file.attrs["author_email"] == "wdudley@farama.org" + str(n_data - 1)
```
younik, line 279: btw, this behavior is a little weird in combine dataset (you just save the attributes of the last)
Anyway, it is fixed in https://github.com/Farama-Foundation/Minari/pull/133 that we will merge after, so it is okay for now

younik (changes requested):

tests/utils/test_dataset_combine.py
im-Kitsch: Ah, yes, I don't think only saving the attribute of last is a good choice, but it's just to keep same with `combine_dataset()` implementation. Definitely it will be changed in the future, but this PR is just to save the fogortten metadata `code_permalink` and `algorithm_name`.

younik (approved):
Just fix the pre-commit, and then looks good to me

im-Kitsch:
> Just fix the pre-commit, and then looks good to me

Hi, I think the pre-commit is fixed now.

PR #160, opened 2023-11-10
Run tests in a temporary Minari dataset dir

### LINKED ISSUE #154: [Proposal] Tests should not leave leftovers
### Proposal

Always remove the data generated by tests, whenever they fail or not.

### Motivation

Currently, tests leave leftovers. This can make all the tests fail after one failed, because of `minari.list_local_datasets()`

### Pitch

We can use `pytest.fixture` and a temporary directory to be sure everything is cleaned after. 
See [here](https://github.com/Farama-Foundation/Minari/blob/707f138daf9641c3d697a7f0c88b1485e4d920f2/tests/dataset/test_minari_storage.py#L10C3-L10C3) for an example.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Minari/issues) in the repo (**required**)

alexdavey:
# Description

Fixes #154. Currently, tests create datasets in the main dataset directory (`~/.minari/datasets`), deleting them at the end. If a test fails half-way, or is interrupted, this can leave leftover files.

This PR patches the `MINARI_DATASETS_PATH` environment variable during test time so that each test dataset is stored in a separate temporary directory, which is cleaned up even if the test fails.

To check the fix works as intended, I have added a "test" `test_create_leftover` that creates an empty directory, emulating the leftover of a failed test. Without the environment variable fix in `conftest.py`, this will cause subsequent tests to fail.

## Type of change

# Checklist:


rodrigodelazcano:
This LGTM. Thanks! @younik any suggestions before merging?

tests/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+import tempfile
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def tmp_dataset_dir():
+    """Generate a temporary directory for Minari datasets."""
+    tmp_dir = tempfile.TemporaryDirectory()
+    os.environ["MINARI_DATASETS_PATH"] = tmp_dir.name
```
younik, line 11: should you set `MINARI_DATASETS_PATH` back to the original value after it?

I guess if you interrupt tests for some reasons, there is a risk this will remain written anyway, but c'est la vie, I don't see a better way to do it

tests/test_misc.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def test_create_leftover():
+    """Create a leftover dataset to simulate a failed test, to verify test isolation."""
+    dataset_dir = os.environ.get(
+        "MINARI_DATASETS_PATH",
+        os.path.join(os.path.expanduser("~"), ".minari/datasets/"),
+    )
+    new_dataset_path = Path(dataset_dir) / "leftover_dataset/data"
+    new_dataset_path.mkdir(parents=True)
```
younik, line 12: I would remove this test before merging as we don't really need to test for it after

younik (changes requested):
This is very good and clean indeed, many thanks @alexdavey 
I just added a couple of comments

tests/conftest.py
alexdavey: Good idea, thanks! I have rewritten it to use MonkeyPatch, which automatically undoes the environment variable change upon exiting the with block.

alexdavey:
Thanks @rodrigodelazcano and @younik for the review. I believe I have addressed your comments, let me know if there are any remaining issues.

younik (approved):
That's perfect, thank you!

PR #163, opened 2023-11-11
Add obs/act spaces when combining datasets

rodrigodelazcano:
# Description

Datasets should have the same observation and action space before being combined. The spaces should also be stored as an attribute in case the dataset has different spaces from the env used to create the dataset (then spaces shouldn't be inferred when loading the dataset)

# Checklist:

### DIFF
--- a/minari/utils.py
+++ b/minari/utils.py
@@ -156,13 +156,15 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
     Tests if the datasets were created with the same environment (`env_spec`) and re-calculates the
     `max_episode_steps` argument.
 
+    Also checks that the datasets obs/act spaces are the same.
+
     Args:
         datasets_to_combine (List[MinariDataset]): list of MinariDataset to combine
 
     Returns:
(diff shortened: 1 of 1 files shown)


minari/utils.py
```diff
@@ -185,6 +187,10 @@ def validate_datasets_to_combine(datasets_to_combine: List[MinariDataset]) -> En
         env_spec == combine_env_spec[0] for env_spec in combine_env_spec
     ), "The datasets to be combined have different values for `env_spec` attribute."
 
+    # Check that all datasets have the same action/observation space
+    assert all(dataset.action_space == datasets_to_combine[0].action_space for dataset in datasets_to_combine), "The datasets to combine must have the same action space"
+    assert all(dataset.observation_space == datasets_to_combine[0].observation_space for dataset in datasets_to_combine), "The datasets to combine must have the same observation space"
```
younik, line 192: maybe change these lines to
```python
if any(dataset.action_space != datasets_to_combine[0].action_space for dataset in datasets_to_combine):
   raise ValueError("...")
```
?

younik (approved):
LGTM

minari/utils.py
rodrigodelazcano: Done

PR #164, opened 2023-11-11
Mandatory spaces

rodrigodelazcano:
# Description

Always add the `observation_space` and `action_space` attributes to metadata when creating a new dataset. If this values are not provided during dataset creation they will be inferred from the `env_spec` used to collect the data during dataset creation as well.

# Checklist:


rodrigodelazcano:
Building the docker file is failing due to some issues with Docker Hub https://github.com/docker-library/python/issues/884

minari/data_collector/data_collector.py
```diff
@@ -108,13 +108,8 @@ def __init__(
...
 
-        if observation_space is None:
-            observation_space = self.env.observation_space
-        self.dataset_observation_space = observation_space
-
-        if action_space is None:
-            action_space = self.env.action_space
-        self.dataset_action_space = action_space
+        self.dataset_observation_space = self._storage.observation_space
+        self.dataset_action_space = self._storage.action_space
```
younik, line 112: It looks like self.dataset_observation_space is not even used; maybe we can just remove it?

minari/dataset/minari_storage.py
```diff
@@ -328,31 +336,22 @@ def observation_space(self) -> gym.Space:
...
-                if "observation_space" in file.attrs.keys():
-                    serialized_space = file.attrs["observation_space"]
-                    assert isinstance(serialized_space, str)
-                    self._observation_space = deserialize_space(serialized_space)
-                else:
-                    env_spec_str = file.attrs.get("env_spec")
-                    assert isinstance(env_spec_str, str)
-                    env_spec = EnvSpec.from_json(env_spec_str)
-                    self._observation_space = gym.make(env_spec).observation_space
+                assert "observation_space" in file.attrs.keys(), "Minari hdf5 datasets must contain an observation_space attribute."
```
younik, line 339: This will be a breaking change tho, we should fix all past datasets (?)

younik (changes requested):
Thanks for the PR @rodrigodelazcano 
We should also change the function `create_dataset_from_buffer` accordingly, but then looks to me

minari/data_collector/data_collector.py
rodrigodelazcano: It's used in the `step` method, but true we can just get it from _storage when needed

minari/dataset/minari_storage.py
rodrigodelazcano: I'll need to check but I think that all of the latest versions of the datasets compatible with `minari>=0.4.0` have the observation and action space attributes. I'll double check before we merge this PR

rodrigodelazcano:
> Thanks for the PR @rodrigodelazcano We should also change the function `create_dataset_from_buffer` accordingly, but then looks to me

`create_dataset_from_buffer` also uses `MinariStorage.new()` to generate a new dataset file and add the mandatory space attributes inside this method. No need to change anything inside `create_dataset_from_buffer` I think. I my missing anything @younik ?

younik:
> > Thanks for the PR @rodrigodelazcano We should also change the function `create_dataset_from_buffer` accordingly, but then looks to me
> 
> `create_dataset_from_buffer` also uses `MinariStorage.new()` to generate a new dataset file and add the mandatory space attributes inside this method. No need to change anything inside `create_dataset_from_buffer` I think. Am I missing anything @younik ?

Oh you are right, my bad
Then LGTM, you can merge if all the datasets have the spaces in metadata

younik (approved):

PR #169, opened 2023-11-24
Deprecate create_dataset_from_collector_env

avjmachine:
# Description

- Move the function `minari.create_dataset_from_collector_env() to env.create_dataset(). Add deprecation warning for the same.
- Add deprecation warning for making env.save_to_disk private.
Both changes planned for release 0.5.0

## Type of change

### Screenshots

N/A

# Checklist:

### DIFF
--- a/docs/api/data_collector.md
+++ b/docs/api/data_collector.md
@@ -22,6 +22,7 @@ data_collector_callbacks/episode_metadata_callback
 .. autofunction:: minari.DataCollectorV0.step
 .. autofunction:: minari.DataCollectorV0.reset
 .. autofunction:: minari.DataCollectorV0.close
+.. autofunction:: minari.DataCollectorV0.create_dataset
 .. autofunction:: minari.DataCollectorV0.save_to_disk
 .. autofunction:: minari.DataCollectorV0.clear_buffer_to_tmp_file
 .. autofunction::minari.DataCollectorV0._add_to_episode_buffer
--- a/docs/api/data_collector.md
+++ b/docs/api/data_collector.md
@@ -23,7 +23,6 @@ data_collector_callbacks/episode_metadata_callback
 .. autofunction:: minari.DataCollectorV0.reset
 .. autofunction:: minari.DataCollectorV0.close
 .. autofunction:: minari.DataCollectorV0.create_dataset
-.. autofunction:: minari.DataCollectorV0.save_to_disk
 .. autofunction:: minari.DataCollectorV0.clear_buffer_to_tmp_file
 .. autofunction::minari.DataCollectorV0._add_to_episode_buffer
 ```
--- a/docs/api/minari_functions.md
+++ b/docs/api/minari_functions.md
@@ -10,7 +10,6 @@ minari_dataset/episode_data
 ## Create Minari Dataset
 
 ```{eval-rst}
-.. autofunction:: env.create_dataset
 .. autofunction:: minari.create_dataset_from_buffers
 ```
 
--- a/docs/api/minari_functions.md
+++ b/docs/api/minari_functions.md
@@ -10,7 +10,7 @@ minari_dataset/episode_data
 ## Create Minari Dataset
 
 ```{eval-rst}
-.. autofunction:: minari.create_dataset_from_collector_env
+.. autofunction:: env.create_dataset
 .. autofunction:: minari.create_dataset_from_buffers
 ```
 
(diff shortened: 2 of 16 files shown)


avjmachine:
@younik Faced a circular import issue due to DataCollectorV0 being still used in the create_dataset_from_collector_env argument type, while env.create_data() uses functions from utils.py. 
So, I currently changed the argument type to Any to ensure tests run properly. Another option is to move the required functions and their dependencies to data_collector.py and change imports in all other files. Please suggest if there is a better option than both of these.

docs/api/minari_functions.md
```diff
@@ -10,7 +10,7 @@ minari_dataset/episode_data
 ## Create Minari Dataset
 
 ```{eval-rst}
-.. autofunction:: minari.create_dataset_from_collector_env
+.. autofunction:: env.create_dataset
```
younik, line 13: I don't think this generates the documentation properly
Can you check? You can build the doc locally

docs/api/data_collector.md
```diff
@@ -22,6 +22,7 @@ data_collector_callbacks/episode_metadata_callback
 .. autofunction:: minari.DataCollectorV0.step
 .. autofunction:: minari.DataCollectorV0.reset
 .. autofunction:: minari.DataCollectorV0.close
+.. autofunction:: minari.DataCollectorV0.create_dataset
 .. autofunction:: minari.DataCollectorV0.save_to_disk
```
younik, line 26: Can you already `save_to_disk` from doc?

docs/content/basic_usage.md
```diff
@@ -52,7 +52,7 @@ In this example, the :class:`minari.DataCollectorV0` wraps the `'CartPole-v1'` e
 ```{eval-rst}
 To create a Minari dataset first we need to step the environment with a given policy to allow the :class:`minari.DataCollectorV0` to record the data that will comprise the dataset. This is as simple as just looping through the Gymansium MDP API. For our example we will loop through ``100`` episodes of the ``'CartPole-v1'`` environment with a random policy.
 
-Finally, we need to create the Minari dataset and give it a name id. This is done by calling the :func:`minari.create_dataset_from_collector_env` Minari function which will move the temporary data recorded in the :class:`minari.DataCollectorV0` environment to a permanent location in the `local Minari root path </content/dataset_standards>`_ with the Minari dataset standard structure.
+Finally, we need to create the Minari dataset and give it a name id. This is done by calling the :func:`env.create_dataset` Minari function which will move the temporary data recorded in the :class:`minari.DataCollectorV0` environment to a permanent location in the `local Minari root path </content/dataset_standards>`_ with the Minari dataset standard structure.
```
younik, line 55: In the doc, maybe you should write `DataCollectorV0.create_dataset`? As env can be unclear what it means

younik: [there are multiple places where this comment is valid]

docs/tutorials/dataset_creation/point_maze_dataset.py
```diff
@@ -396,8 +395,7 @@ def __call__(
 
     obs, rew, terminated, truncated, info = collector_env.step(action)
 
-dataset = minari.create_dataset_from_collector_env(
-    collector_env=collector_env,
+dataset = env.create_dataset(
```
younik, line 398: to fix tests, `env` -> `collector_env` here

minari/utils.py
```diff
@@ -567,7 +566,7 @@ def create_dataset_from_buffers(
 
 def create_dataset_from_collector_env(
     dataset_id: str,
-    collector_env: DataCollectorV0,
+    collector_env: Any,
```
younik, line 569: to avoid circular import,  in DataCollectorV0, you can import `from minari.utils import _generate_dataset_metadata, _generate_dataset_path` inside the `create_dataset` function (please also add a #TODO comment saying to move it on top with 0.5.0 release)

younik (changes requested):
Very good thanks, I suggested some minor changes

docs/api/minari_functions.md
avjmachine: Couldn't spot it since I had failures in building the docs. Now removed it since it is no longer required due to the deprecation.

docs/api/data_collector.md
avjmachine: done

docs/content/basic_usage.md
avjmachine: Modified in all the instances in the docs.

docs/tutorials/dataset_creation/point_maze_dataset.py
avjmachine: done

minari/utils.py
avjmachine: This change is also done.

younik (approved):
LGTM, thanks!

PR #172, opened 2023-12-01
Add automatic seeding of Datacollector reset

alexdavey:
# Description

This PR adds automatic seeding of environments by the DataCollector. If no explicit seed is set, one will be automatically generated when `DataCollector.reset()` is called. The seed is a `uint64`, the maximum integer supported by H5Py.

Reproducibility tests have also been added, both for when an explicit seed is set (addressing issue #15) and when autoseeding.

Other changes:
- Dummy envs are now deterministic
- Minor bugfix: `seed=0` was previously recorded as `seed=None`.

## Type of change

# Checklist:

### DIFF
--- a/tests/data_collector/test_data_collector.py
+++ b/tests/data_collector/test_data_collector.py
@@ -6,6 +6,8 @@ from minari import DataCollector, EpisodeData, MinariDataset, StepDataCallback
 from tests.common import check_load_and_delete_dataset, register_dummy_envs
 
 
+MAX_UINT64 = np.iinfo(np.uint64).max
+
 register_dummy_envs()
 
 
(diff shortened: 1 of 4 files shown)


tests/data_collector/test_data_collector.py
```diff
@@ -141,3 +143,54 @@ def test_truncation_without_reset(dataset_id, env_id):
...
+
+        obs, _ = env.reset(seed=episode.seed)
+        env.action_space.seed(policy_seed)
+
+        assert np.allclose(obs, episode.observations[0])
+
+        for k in range(episode.total_timesteps):
+            obs, rew, _, _, _ = env.step(env.action_space.sample())
+            assert np.allclose(obs, episode.observations[k + 1])
+            assert rew == episode.rewards[k]
```
younik, line 194: can you change this loop to:
1) Iterate over the dataset instead of sampling
2) Instead of using `policy_seed` and `env.action_space.sample()`, use the action from the dataset
3) Add also the checks for `termination` and `truncations`

younik (commented):
Very good, thank you @alexdavey! I just added a comment

alexdavey:
Thank you for the review @younik! I believe I have addressed all of the comments.

tests/data_collector/test_data_collector.py
```diff
@@ -176,21 +174,22 @@ def test_reproducibility(seed):
...
 
         assert np.allclose(obs, episode.observations[0])
 
         for k in range(episode.total_timesteps):
-            obs, rew, _, _, _ = env.step(env.action_space.sample())
+            obs, rew, term, trunc, _ = env.step(episode.actions[k])
             assert np.allclose(obs, episode.observations[k + 1])
             assert rew == episode.rewards[k]
+            assert not term
+            assert trunc == (k == episode.total_timesteps - 1)
```
younik, line 193: does something like
```python
assert term == episode.terminations[k]
assert trunc == episode.truncations[k]
```

works?

alexdavey: Sorry, it does of course. Thanks for spoting this!

younik: No worries, thanks indeed, looks good!

PR #180, opened 2024-01-07
Add IQL-TorchRL tutorial

alexdavey:
# Description

This PR adds a tutorial showcasing Minari's integration with TorchRL: training an agent with Implicit Q-Learning on the `pen-human-v1` dataset.

# Checklist:


docs/tutorials/using_datasets/IQL_torchrl.py
```diff
@@ -0,0 +1,522 @@
...
+# We can compute the cumulative reward by summing ``tensordict['next', 'reward']``:
+
+# %%
+print(f"Cumulative reward: {tensordict['next', 'reward'].sum():.2f}")
+
+# %%
+# .. code-block:: text
+#
+#     Cumulative reward: -31.10
+#
```
younik, line 133: I get a different value when running it (-871.17)

docs/tutorials/using_datasets/IQL_torchrl.py
```diff
@@ -0,0 +1,522 @@
...
+# .. math::
+#   L_{\rm TD}(\theta) = \mathbb{E}_{(s, a, s') \sim \mathcal{D}} \left[ \left( r(s, a) + \gamma \max_{a'} Q_{\hat{\theta}}(s', a') - Q_\theta(s, a) \right)^2 \right]
+#
+#
+# .. figure:: /_static/img/tutorials/IQL_torchrl_expectile_example.png
+#    :width: 250
+#    :figwidth: 300
+#    :align: right
+#
+#    Expectiles of an example conditional distribution :math:`y \sim m_\tau(s)`. For :math:`\tau = 0.5` the expectile is the mean while for :math:`\tau \approx 1` it approximates the maximum of the distribution `(Kostrikov et al 2021) <https://arxiv.org/abs/2110.06169>`_.
```
younik, line 204: Will this be displayed in this position? It looks like it breaks the flow

docs/tutorials/using_datasets/IQL_torchrl.py
```diff
@@ -0,0 +1,522 @@
...
+# The training loop is essentially the standard PyTorch gradient descent loop:
+#
+# 1. Sample a batch of transitions from the dataset :math:`\mathcal{D}`.
+# 2. Compute the loss :math:`\ell = L_V(\psi) + L_Q(\theta) + L_\pi(\phi)`.
+# 3. Backpropagate the gradients and update the networks, including the target Q-network.
+
+# %%
+from tqdm.auto import tqdm
+
+iterations = 50_000
```
younik, line 409: the loss plot seems to converge after a few iterations (the dataset is very small)
Also, the return is not clear if it is benefiting from the "long" training.

Maybe we can reduce the number of iterations or use the "pen-expert-v1" which is larger

docs/tutorials/using_datasets/IQL_torchrl.py
```diff
@@ -0,0 +1,522 @@
...
+# Evaluated over 100 episodes, the final performance is:
+
+# %%
+final_score = evaluate_policy(env, model[0], num_eval_episodes=100)
+print(f"Cumulative reward (averaged over 100 episodes): {final_score:.2f}")
+
+# %%
+# .. code-block:: text
+#
+#     Cumulative reward (averaged over 100 episodes): 1952.30
```
younik, line 479: as before, I get another value. Maybe, you run some cells more than one time (?)
Before merging we should run the whole notebook and fill in the correct values (and check reproducibility)

younik (commented):
Great tutorial, thank you!

docs/tutorials/using_datasets/IQL_torchrl.py
alexdavey: It turns out I wasn't seeding everything. Thanks for spotting this!

docs/tutorials/using_datasets/IQL_torchrl.py
alexdavey: When rendered it floats to the right so it doesn't disrupt the text. However, if you have a better position in mind I would be happy to move it.

docs/tutorials/using_datasets/IQL_torchrl.py
alexdavey: To keep things fast, it only uses a small number of rollouts (20) to estimate the return, so the return plot is rather noisy. I haven't tested it in detail but 50K iterations seems to be the amount required to get a decent return across a range of seeds. However I decreased the number of iterations to 10K so that it doesn't timeout the tests.

docs/tutorials/using_datasets/IQL_torchrl.py
alexdavey: Hopefully the seeding fix has addressed this as well. If not let me know!

alexdavey:
Thanks for the review @younik! I believe I have addressed all of the comments, if there any unresolved issues please let me know.

younik (approved):
LGTM, thanks!

PR #181, opened 2024-01-12
Revert dataset spec total_steps type to int

alexdavey:
# Description

A [recent commit](https://github.com/Farama-Foundation/Minari/commit/dd8406ee8dce27162724fced1f7147d702a70027) modified the type of `MinariDatasetSpec.total_steps` from `int` to `np.int64` (source [here](https://github.com/Farama-Foundation/Minari/blob/e24113bf09b5b1769dafbaf6bb0b2c9b8eb9ffbe/minari/dataset/minari_dataset.py#L54)). This breaks the [TorchRL integration](https://github.com/pytorch/rl/pull/1721) as `np.int64` is not a natively JSON-encodable type. This PR reverts the type back to `int`.

One can confirm that the PR fixes the change with the following script:
```python
# Clear any cached files first: rm -r ~/.cache/torchrl/minari
from torchrl.data.datasets.minari_data import MinariExperienceReplay
data = MinariExperienceReplay("door-human-v1", batch_size=32)
```
With the current version of Minari built from source, this raises `TypeError: Object of type int64 is not JSON serializable`. 

Note: I opted to not make this a proper test, since it requires torch-nightly, which is platform-dependent and probably too cumbersome of a dependency to add to the tests.

## Type of change

# Checklist:

### DIFF
--- a/minari/dataset/minari_dataset.py
+++ b/minari/dataset/minari_dataset.py
@@ -51,7 +51,7 @@ def parse_dataset_id(dataset_id: str) -> tuple[str | None, str, int]:
 class MinariDatasetSpec:
     env_spec: Optional[EnvSpec]
     total_episodes: int
-    total_steps: np.int64
+    total_steps: int
     dataset_id: str
     combined_datasets: List[str]
     observation_space: gym.Space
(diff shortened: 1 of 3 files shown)


younik (approved):
LGTM, thanks!

PR #194, opened 2024-03-04
Add description arg

enerrio:
# Description

This PR adds an optional description arg to the `create_dataset` and `create_dataset_from_buffers` functions which allow a user to include a brief description of the dataset they are creating. If a description is not provided then a warning will be displayed to the user stating it is recommended to include one.

A minimal working example displaying usage of description:
```python
import gymnasium as gym
from minari.data_collector import DataCollector

env = gym.make("Blackjack-v1")
env = DataCollector(env)

for episode in range(10):
    done = False
    _, info = env.reset()  # {}
    while not done:
        action = env.action_space.sample()
        _, _, terminated, truncated, _ = env.step(action)
        done = terminated or truncated

dataset = env.create_dataset(
    dataset_id="blackjack-random-v1",
    algorithm_name="Random-Policy",
    code_permalink="dummy-permalink",
    author="Farama",
    author_email="dummy-email",
    minari_version="==0.4.3",
    description="A random policy for 10 fixed episodes" # commenting out this line will result in a warning to user
)
```

## Type of change

### Screenshots

N/A

# Checklist:

### DIFF
--- a/minari/data_collector/data_collector.py
+++ b/minari/data_collector/data_collector.py
@@ -319,6 +319,7 @@ class DataCollector(gym.Wrapper):
         expert_policy: Optional[Callable[[ObsType], ActType]] = None,
         num_episodes_average_score: int = 100,
         minari_version: Optional[str] = None,
+        description: Optional[str] = None,
     ):
         """Create a Minari dataset using the data collected from stepping with a Gymnasium environment wrapped with a `DataCollector` Minari wrapper.
 
@@ -341,6 +342,7 @@ class DataCollector(gym.Wrapper):
                                                                             `ref_max_score` and `expert_policy` can't be passed at the same time. Default to None
             num_episodes_average_score (int): number of episodes to average over the returns to compute `ref_min_score` and `ref_max_score`. Default to 100.
             minari_version (Optional[str], optional): Minari version specifier compatible with the dataset. If None (default) use the installed Minari version.
+            description (Optional[str], optional): description of the dataset being created. Defaults to None.
 
         Returns:
             MinariDataset
@@ -359,6 +361,7 @@ class DataCollector(gym.Wrapper):
             expert_policy,
             num_episodes_average_score,
             minari_version,
+            description,
         )
 
         self._save_to_disk(dataset_path, metadata)
--- a/minari/utils.py
+++ b/minari/utils.py
@@ -347,6 +347,7 @@ def _generate_dataset_metadata(
     expert_policy: Optional[Callable[[ObsType], ActType]],
     num_episodes_average_score: int,
     minari_version: Optional[str],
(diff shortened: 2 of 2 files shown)


minari/utils.py
```diff
@@ -385,6 +386,15 @@ def _generate_dataset_metadata(
     else:
         dataset_metadata["algorithm_name"] = algorithm_name
 
+    if description is None:
+        print("here")
```
younik, line 390: Can you remove this print?

minari/data_collector/data_collector.py
```diff
@@ -341,6 +342,7 @@ def create_dataset(
                                                                             `ref_max_score` and `expert_policy` can't be passed at the same time. Default to None
             num_episodes_average_score (int): number of episodes to average over the returns to compute `ref_min_score` and `ref_max_score`. Default to 100.
             minari_version (Optional[str], optional): Minari version specifier compatible with the dataset. If None (default) use the installed Minari version.
+            description (Optional[str], optional): description of the dataset being created. Defaults to None.
```
younik, line 345: Actually, we should use `(Optional[str]l)`, also the line above is wrong; can you fix it?

minari/utils.py
```diff
@@ -506,6 +517,7 @@ def create_dataset_from_buffers(
         action_space (Optional[gym.spaces.Space]): action space of the environment. If None (default) use the environment action space.
         observation_space (Optional[gym.spaces.Space]): observation space of the environment. If None (default) use the environment observation space.
         minari_version (Optional[str], optional): Minari version specifier compatible with the dataset. If None (default) use the installed Minari version.
+        description (Optional[str], optional): description of the dataset being created. Defaults to None.
```
younik, line 520: same as other comment

younik (changes requested):
Thank you @enerrio!
I just added a couple of comments

minari/utils.py
enerrio: oop. fixed!

minari/data_collector/data_collector.py
enerrio: Do you mean wanting the description to be required instead of defaulting to None?

younik: No, it is just for the documentation `Optional[str]l` means it is optional already, so we can remove `, optional`

enerrio: I see throughout the repo many of the other optional args also have the same format of `(Optional[<dtype>], optional)` in docstrings. Should these be kept for consistency?

younik: Yes, actually, we should standardize it across all the documentation, so let's do it in another PR.

younik (approved):

PR #201, opened 2024-04-15
Fix download flags and update CLI

### LINKED ISSUE #200: [Bug Report] listed datasets are unavailable for a given version - why list them?
This is more a question than a bug report.
In torchrl we test a random set of datasets extracted from `list_remote_datasets`
Since the last release, this test has been broken: https://github.com/pytorch/rl/actions/runs/8649362912/job/23715408536

The reason is that a dataset picked from the list has the [right major version](https://github.com/pytorch/rl/blob/0d007483317b9b1f7298082f1d40401b971220c3/test/test_libs.py#L2560-L2566) (0.4) but the wrong minor, and isn't available for this minor anymore.

My questions are:
- If I want to pick up datasets that work with my version of Minari, how can I do this (and not pick datasets that don't)?
- Why keeping in the list datasets that don't work with my version of minary? If it's for backward referencing, maybe a flag in `list_remote_datasets` to display non-compatible datasets would be useful (I'd expect people to be surprised to see datasets they can't work with in the list except if they ask for them)
- I might be wrong but it seems to me that the dataset that is causing our error (pointmaze-large-dense-v1) has changed in between two minor versions. IMO minor versions should not introduce any change of API unless a bug is solved on the user side. In other words, whatever worked in one major should work in all versions of that major (I understand not all libs abide by this rule but ideally I think it should be the case)

alexdavey:
# Description

When downloading a dataset, one of the flags in the call to `get_remote_dataset_versions` is incorrectly set, and loads a list of the latest versions rather than compatible versions. This prevents users from using older dataset versions even if they are compatible (unless they force-download first). This fixes #200. Tests have been added to check the warning/error logic in `download_dataset()`.

To make the version availability more transparent, the `minari list` CLI has been updated to show a list of compatible versions for each dataset:

Screenshot 2024-04-17 at 21 02 24

This PR also includes some fixes for issues in `minari show` when rendering `">="` and `serialized` spaces.

## Type of change

# Checklist:

### DIFF
--- a/minari/cli.py
+++ b/minari/cli.py
@@ -1,6 +1,7 @@
 """Minari CLI commands."""
 import os
-from typing import List, Optional
+from collections import defaultdict
+from typing import Any, Dict, List, Optional
 from typing_extensions import Annotated
 
 import typer
@@ -12,6 +13,7 @@ from rich.text import Text
 from rich.tree import Tree
 
 from minari import __version__
+from minari.dataset.minari_dataset import parse_dataset_id
 from minari.storage import get_dataset_path, hosting, local
 from minari.utils import combine_datasets, get_dataset_spec_dict, get_env_spec_dict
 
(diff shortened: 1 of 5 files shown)


younik (approved):
LGTM, thank you for this!

PR #203, opened 2024-04-18
Fix Black

alexdavey:
# Description

The `force-exclude` setting for the Black configuration in `pyproject.toml` is a regex which was incorrectly set, matching and excluding a significant proportion of the codebase. This PR fixes that setting. The changes outside `pyproject.toml` are due to Black formatting.

## Type of change


younik (approved):
aha, good catch, thank you!

PR #218, opened 2024-07-12
Moved h5py, google_cloud_storage, pyarrow and tqdm to optional depend…

JosephCarrino:
Updated dependency list removing "portion" and "packaging", which are not used.
Some dependencies (i.e., h5py, google-cloud-storage) are now optional and not mandatory anymore, since they are used only for specific features. An `ImportError` expection is raised in case one is needed and not installed. Also, tqdm is now in `docs/requirements.txt`.


pyproject.toml
```diff
@@ -26,26 +26,27 @@ classifiers = [
...
+all = ["minari[pyarrow,h5py,gcs]"]
+
 testing = [
     "pytest>=7.1.3",
     "gymnasium-robotics>=1.2.3",
     "mktestdocs",
     "torchrl",
-    "agilerl"
+    "agilerl",
+    "h5py>=3.8.0",
```
younik, line 49: this should not be needed (as you already install "all" for testing)

minari/storage/remotes/gcp.py
```diff
@@ -2,12 +2,18 @@
...
 from typing import Any, Optional
 
-from google.cloud import storage
-from tqdm.auto import tqdm
-
 from minari.storage.remotes.cloud_storage import CloudStorage
 
 
+try:
+    from google.cloud import storage
```
younik, line 9: `import google.cloud.storage as gcp_storage` should solve the pre-commit issue (and you need to change the storage reference accordingly)

younik (changes requested):
Thank you for the PR @JosephCarrino!

It looks good, I just added a couple of comments. I will also have a deep look for Gymnasium dependency soon.

younik:
Oh also for Test Documentation, you should change this line to also install all: 
https://github.com/Farama-Foundation/Minari/blob/338b19a8ce0838d188fc422f4d5de49bca47ed65/.github/workflows/docs-test.yml#L25

Then, you don't need anymore tqdm in docs/requirements.txt

minari/dataset/_storages/arrow_storage.py
```diff
@@ -7,16 +7,21 @@
...
+    raise ImportError(
+        'pyarrow is not installed. Please install it using `pip install "minari[arrow]"`'
+    )
 
 from minari.data_collector.episode_buffer import EpisodeBuffer
 from minari.dataset.minari_storage import MinariStorage
 
 
 class ArrowStorage(MinariStorage):
-    FORMAT = "arrow"
```
younik, line 18: Tests fails because there are places where FORMAT is used; let's keep FORMAT.

It would be nice to assert in __init__ that the key is the same as MinariStorage.FORMAT. If you have time, change the __init__ to be similar to the of CloudStorage: use a function `get_minari_storage(key: str)` (that uses the private registry of callables) and assert that `storage.FORMAT == key`. (Then you will also need to change the places where the registry is used to use the function instead). This is a very minor problem, so it's okay to skip it.

younik (changes requested):
I left a couple of comments, then it should be good I believe 🤞

minari/dataset/_storages/arrow_storage.py
JosephCarrino: I should have implemented this too, however this is a bite more tricky than the CloudStorage one, since this registry is used in many tests to get the list of the available storages. What I did is creating also a function `get_storage_keys() -> List[str]` that preserve the fact of registry being private and returns the list of its keys. I replaced the previous `_storages.registry.keys()` which was in some tests and functions by this new function.

younik: Sounds good, can you push the changes? @JosephCarrino

younik (approved):
LGTM, thanks!

PR #226, opened 2024-08-05
ENH: add requirements in metadata

younik:
Add requirements in metadata that can be used to indicate which packages and versions are needed for the environment to work as expected.

### DIFF
--- a/docs/_scripts/generate_gif.py
+++ b/docs/_scripts/generate_gif.py
@@ -36,7 +36,7 @@ def generate_gif(dataset_id, num_frames=256, fps=16):
             act = _space_at(episode.actions, step_id)
             env.step(act)
             images.append(env.render())
-        
+
         episode_id += 1
 
     env.close()
--- a/docs/_scripts/generate_gif.py
+++ b/docs/_scripts/generate_gif.py
@@ -1,4 +1,6 @@
 import os
+import subprocess
+import sys
 
 import imageio
 
@@ -16,6 +18,10 @@ def _space_at(values, index):
 
 def generate_gif(dataset_id, num_frames=256, fps=16):
     dataset = minari.load_dataset(dataset_id, download=True)
+    requirements = dataset.storage.metadata.get("requirements", [])
+    for req in requirements:
+        subprocess.check_call([sys.executable, "-m", "pip", "install", f'"{req}"'])
+
     env = dataset.recover_environment(render_mode="rgb_array")
     images = []
     ep_id = 0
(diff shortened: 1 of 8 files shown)


minari/data_collector/data_collector.py
```diff
@@ -259,6 +260,7 @@ def create_dataset(
                                                                             `ref_max_score` and `expert_policy` can't be passed at the same time. Default to None
             num_episodes_average_score (int): number of episodes to average over the returns to compute `ref_min_score` and `ref_max_score`. Default to 100.
             description (str, optional): description of the dataset being created. Defaults to None.
+            requirements (list of str, optional): list of requirements to load the environment and reproduce the dataset. Defaults to None.
```
alexdavey, line 263: It may be good to add some very brief documentation about the requirements format used. For example, it may not be obvious that a full pip-style requirement specification is supported.

minari/dataset/minari_dataset.py
```diff
@@ -162,12 +165,34 @@ def recover_environment(self, eval_env: bool = False, **kwargs) -> gym.Env:
         """Recover the Gymnasium environment used to create the dataset.
 
         Args:
-            eval_env (bool): if True the returned Gymnasium environment will be that intended to be used for evaluation. If no eval_env was specified when creating the dataset, the returned environment will be the same as the one used for creating the dataset. Default False.
+            eval_env (bool): if True, the returned Gymnasium environment will be that intended to be used for evaluation. If no eval_env was specified when creating the dataset, the returned environment will be the same as the one used for creating the dataset. Default False.
+            install_requirements (bool): if True, eventual requirements will be installed before creating the environment. Default False.
```
alexdavey, line 169: Is the word "eventual" here a typo?

alexdavey (commented):
Thanks @younik! Looks good, just a few very minor comments.

minari/dataset/minari_dataset.py
younik: Oh, I actually removed this flag, leaving only the warnings, so I remove this line, thanks!

docs/_scripts/generate_gif.py
```diff
@@ -16,23 +18,29 @@ def _space_at(values, index):
...
         for step_id in range(episode.total_steps):
             act = _space_at(episode.actions, step_id)
             env.step(act)
             images.append(env.render())
+        
+        episode_id += 1
 
-        ep_id += 1
+    env.close()
+    minari.delete_dataset(dataset_id)
```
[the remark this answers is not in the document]
younik: Make sense; then also downloading is a bit weird for a function that should just generate the gif. I change it to assume that the dataset is in local datasets and I move downloading and deleting outside.

minari/data_collector/data_collector.py
younik: Right, added!

PR #275, opened 2025-02-14
JPEG encoding and decoding if the observation is an image

gabrielemaraglino:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

### Screenshots

# Checklist:


younik (changes requested):
Thanks for adding this feature!

Can you add the line `.idea` in the `.gitignore` file?
Also, can you add some tests? You can make a fake env that generates images as obs, and check that everything is okay (serialization works and deserialization returns the same obs). Check the dataset_creation tests for the examples

tests/test_jpeg_serialization.py
```diff
@@ -0,0 +1,97 @@
...
+        }
+        metadata_path = tmp_path.joinpath(METADATA_FILE_NAME)
+        metadata_path.write_text(json.dumps(default_metadata))
+
+        from minari.dataset._storages.hdf5_storage import HDF5Storage
+        storage = HDF5Storage._create(tmp_path, env.observation_space, env.action_space)
+        storage.update_episodes([episode])
+        loaded_episodes = list(storage.get_episodes([0]))
+        loaded_obs = loaded_episodes[0]["observations"]
+        np.testing.assert_array_equal(episode.observations, loaded_obs)
```
younik, line 93: These tests can be merged in a single function, and use pytest.mark.parametrize to test both storages.

Also, in the end, I believe that if you add your test env in the list of envs as mentioned above, you can rely on previous written tests that should already check equality, and this file can be removed

minari/dataset/_storages/arrow_storage.py
```diff
@@ -178,11 +180,20 @@ def _encode_space(space: gym.Space, values: Any, pad: int = 0):
         return pa.StructArray.from_arrays(arrays, names=names)
     elif isinstance(space, _FIXEDLIST_SPACES):
         values = np.asarray(values)
-        assert values.shape[1:] == space.shape
-        values = values.reshape(values.shape[0], -1)
-        values = np.pad(values, ((0, pad), (0, 0)))
-        dtype = pa.list_(pa.from_numpy_dtype(space.dtype), list_size=values.shape[1])
-        return pa.FixedSizeListArray.from_arrays(values.reshape(-1), type=dtype)
+        if values.shape == (4, 84, 84) and values.dtype == np.uint8: # check for image observation (4 stacked greyscale images)
```
younik, line 183: we should not constrain ourselves with 84 x 84 images, but we should accept any size.

Same reasoning for the 4 dimension.

I think the discriminant here is to have Box, with at least 2 dim, and with type uint8. Then you can put a logging.warn saying you are considering it as image, and if this is not intended to disable it via a flag "image_observation" which is defaulted to None. Then you can compute the value of the flag during init (and warn just once). I will clarify later in our meeting.

minari/dataset/_storages/arrow_storage.py
```diff
@@ -207,8 +218,15 @@ def _decode_space(space, values: pa.Array):
             ]
         )
     elif isinstance(space, _FIXEDLIST_SPACES):
-        data = np.stack(values.to_numpy(zero_copy_only=False))
-        return data.reshape(-1, *space.shape)
+        if values.type == pa.binary():  # check for binary data (JPEG)
+            jpeg_images = []
+            for jpeg_bytes in values:
+                image = Image.open(io.BytesIO(jpeg_bytes)).convert("L") # decode JPEG and convert to greyscale
```
younik, line 224: we should work with non-grayscale images as well.

younik (changes requested):

younik (approved):

PR #278, opened 2025-02-23
Fix incorrect path name on windows

### LINKED ISSUE #277: [Bug Report] Incorrect path name on Windows
**Describe the bug**
When downloading a dataset from Hugging Face on Windows, the dataset retrieval fails due to an incorrect file path format. Windows’ native file path construction (e.g., using os.path.join) produces paths with backslashes (\), but the dataset URL and entry points on Hugging Face expect Linux-style paths with forward slashes (/). As a result, the improperly constructed URL leads to an "Entry Not Found" error.

**Details:**

Path Separator Discrepancy:
Windows uses backslashes by default, so using functions like os.path.join(path_in_repo, "data", METADATA_FILE_NAME) produces paths with \ (e.g., .../data\metadata.json). This is problematic in contexts where a URL or a Linux-style path is expected.

Impact on Dataset Download:
The dataset URL constructed with backslashes (or their URL-encoded equivalent) does not match the expected format on Hugging Face. Consequently, the download process fails with an error indicating that the entry is not found.

**Lines In Source Code**
line 154 and line 182 in storage/remotes/huggingface.py
a quick work around is to use "os.path.join(path_in_repo, "data", METADATA_FILE_NAME).replace("\\", "/")"

**System Info**
Windows

**Additional context**

FlutteryEmbers:
# Description

Fix bug where, on Windows, the dataset URL is constructed using a file path generated by os.path.join. This function produces Windows-style paths with backslashes (\), which—when used in a URL—get URL-encoded (e.g., %5C) and result in an "Entry Not Found" error on Hugging Face. Additionally, when parsing the dataset_id from local dataset (local_dsts = local.list_local_datasets() ), the Windows path produces a double slash (\\), leading to an invalid dataset id.

Fixes #277

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/minari/dataset/minari_dataset.py
+++ b/minari/dataset/minari_dataset.py
@@ -36,7 +36,6 @@ def parse_dataset_id(dataset_id: str) -> tuple[str | None, str, int]:
     Raises:
         Error: If the dataset id is not valid dataset regex
     """
-    dataset_id = dataset_id.replace("\\", "/")
     match = DATASET_ID_RE.fullmatch(dataset_id)
     if not match:
         raise ValueError(
--- a/minari/dataset/minari_dataset.py
+++ b/minari/dataset/minari_dataset.py
@@ -36,6 +36,7 @@ def parse_dataset_id(dataset_id: str) -> tuple[str | None, str, int]:
     Raises:
         Error: If the dataset id is not valid dataset regex
     """
+    dataset_id = dataset_id.replace("\\", "/")
     match = DATASET_ID_RE.fullmatch(dataset_id)
     if not match:
         raise ValueError(
--- a/minari/storage/local.py
+++ b/minari/storage/local.py
@@ -87,7 +87,7 @@ def list_local_datasets(
             return
 
         for dir_name in list_non_hidden_dirs(parent_dir):
-            namespaced_dir_name = os.path.join(namespace, dir_name)
+            namespaced_dir_name = pathlib.Path(namespace, dir_name).as_posix()
             recurse_directories(base_path, namespaced_dir_name)
 
     recurse_directories(datasets_path, prefix or "")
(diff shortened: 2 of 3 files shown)


minari/dataset/minari_dataset.py
```diff
@@ -36,6 +36,7 @@ def parse_dataset_id(dataset_id: str) -> tuple[str | None, str, int]:
     Raises:
         Error: If the dataset id is not valid dataset regex
     """
+    dataset_id = dataset_id.replace("\\", "/")
```
younik, line 39: This shouldn't be necessary, isn't it? 
The `dataset_id` should already be in the correct format. If this is not the case, you should change fix the caller

younik (changes requested):

minari/dataset/minari_dataset.py
FlutteryEmbers: I've noticed that the dataset_id is not correct when calling local.list_local_datasets(). The issue appears to be in lines 90 and 99, where os.join is used instead of Path(). Would you recommend that I open a new PR to address this? I'm still new to contributing, so any guidance on what else I should include in the PR would be greatly appreciated.

younik: No worries, I really appreciate you taking the time to contribute!
What I was suggesting is to revert the above change (remove line 39 here), and replace os.join with Path where needed instead.

FlutteryEmbers: May I open a new PR for it? Reverting sometimes can cause problems and leads to unnecessary comments in git history.

younik: Don't worry about git history, as I will squash merge.
You can just delete line 39, it will not cause any problem

FlutteryEmbers: I have just implemented the changes. Additionally, I noticed a few other instances where os.path is used, but since I haven't encountered any issues with them, I decided to leave those as they are.

younik (approved):
LGMT, thanks!

PR #285, opened 2025-04-06
Add support for parquet storage

rob-pitkin:
# Description

Minari currently supports HDF5 and PyArrow formats, we would also like to support Parquet.
The ParquetStorage class is a subclass of the ArrowStorage class with `FORMAT` changed to `"parquet"`

## Type of change

## Benchmarks:
[time_storage.pdf](https://github.com/user-attachments/files/19726430/time_storage.pdf)

# Checklist:


minari/dataset/_storages/parquet_storage.py
```diff
@@ -0,0 +1,295 @@
...
+                value = value.reshape(len(value), *data_shape)
+            nested_dict[field.name] = value
+    return nested_dict
+
+
+class NumpyEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+        return super().default(obj)
```
younik, line 295: I guess this file is mostly a copy paste of arrow_storage.py, isn't it?

I would subclass arrow_storage instead, overriding the FORMAT attribute.

younik (changes requested):
Thanks for the PR!
I just added a comment. 
Can you also perform a benchmark and add the result in the PR comment?
You can use this code: https://gist.github.com/younik/53c8f278dc7c1cac082fe841b74b9532 (you can adapt it to have reasonable timing on your machine, maybe reducing the dataset sizes).

minari/dataset/_storages/parquet_storage.py
```diff
@@ -0,0 +1,22 @@
...
+class ParquetStorage(ArrowStorage):
+    FORMAT = "parquet"
+
+    def __init__(
+        self,
+        data_path: pathlib.Path,
+        observation_space: gym.Space,
+        action_space: gym.Space,
+    ):
+        super().__init__(data_path, observation_space, action_space)
```
younik, line 22: Very small, but I would put this to the arrow_storage file, so you don't need to copy _FIXEDLIST_SPACES

More importantly, did you managed to run the benchmark?

younik (approved):
LGTM, thanks!

PR #328, opened 2026-04-27
Add Python 3.13 to CI and declared support

younik:
# Description

This change adds **Python 3.13** to the declared support surface and runs CI (GitHub Actions and the Docker test image) on **3.13**, while keeping **`requires-python >= 3.8`**.

- **pyproject.toml**: add the `Programming Language :: Python :: 3.13` classifier; keep existing 3.8–3.12 classifiers.
- **CI**: set `actions/setup-python` and Docker `PYTHON_VERSION` to 3.13.
- **Docs**: state support for **Python 3.8 through 3.13** in basic usage.

Branch is based on current **upstream** `main` (`Farama-Foundation/Minari`).

Fixes # (issue), Depends on # (pull request)

## Type of change

> Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

### Screenshots

> Please attach before and after screenshots of the change if applicable.
> To upload images to a PR -- simply drag and drop or copy paste.

# Checklist:

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have run `pytest -v` and no errors are present.
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #334, opened 2026-06-28
Fix CI

younik:
**Move integration and docs tests back to Python 3.12**

In #328 we bumped all CI jobs to Python 3.13, but two jobs started failing: test integrations and test documentation.

It's not really a 3.13 problem in our code. What happens is a dependency mess:

- test integrations: agilerl's latest version needs a torch version that doesn't exist yet, so pip goes back to an old agilerl that pins torch==2.5.1. The torchrl that matches torch 2.5.1 has no 3.13 wheel, so pip installs a newer torchrl built for torch 2.7 instead. That mismatch makes torchrl's C++ part fail to load and the test crashes.
- test documentation: mujoco has no 3.13 wheel for the version we get, so pip tries to build it from source and fails (MUJOCO_PATH not set).

On 3.12 all these packages have matching wheels, so everything installs fine. So for now I just put these two jobs back to 3.12. The other jobs stay on 3.13.

We can move them to 3.13 later once agilerl/torchrl/mujoco sort out their versions.

### DIFF
--- a/.github/workflows/docs-test.yml
+++ b/.github/workflows/docs-test.yml
@@ -16,7 +16,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-            python-version: '3.13'
+            python-version: '3.12'
 
       - name: Install Minari
         run: pip install .[all,testing]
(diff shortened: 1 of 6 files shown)


younik:
Blocked by https://github.com/pytorch/rl/issues/3919

